PR #9, opened 2015-05-25
Added --ssl-protocol=any to phantomjs command

JamesPlayer:
This fixes a bug where phantomjs will not take a screenshot over https.
Details can be found here: http://stackoverflow.com/a/24679134.


freekmurze:
Thanks!

PR #11, opened 2015-08-19
Added ignore SSL errors parameter

martendej:
Fixes possible blank screenshots in PhantomJS (e.g. when using SNI)


freekmurze:
Thanks

PR #17, opened 2015-11-25
Download method

Maxeee09:
return response to directly download browsershot file


freekmurze:
Hi, thanks for your contribution, but I'm not going to merge this.  The responsability of this package is to save a screenshot of an url. Wrapping it into a response is something your application should do.

PR #18, opened 2015-12-02
Add PhantomJS options support

Maxeee09:
Offer the possibility to use differents PhantomJS options


freekmurze:
I like the added options from your first commit, but creating a pdf from an image is not in scope of this package.

Could you create a PR without the pdf related changes?

PR #21, opened 2016-03-11
Fixed issue with rendering complete height

carlcs:
According to http://phantomjs.org/api/webpage/property/viewport-size.html it's necessary to set `height` for this to work.


carlcs:
It's not a good idea to set the height to 9999.

PR #25, opened 2016-08-05
phantomjs setTimeout to onLoadFinish instead

marsd:
Since we are already using resourceTimeout, it's marginally faster to trigger render on `page.onLoadFinished` instead of waiting for a timeout, while `page.onResourceTimeout` will also trigger the same render.

This will have a big impact even on the default timeout (5000) when the page completes loading before reaching the default timeout.


marsd:
Sorry found some bugs in this. Will fix and open another request.

PR #27, opened 2016-08-18
Updgrading PhantomJS version

jamesmills:
PhantomJS 2.1 was released on January 23, 2016. It came with an updated WebKit library based on Qt 5.5.

PhantomJS 2.0 was released on January 23, 2015. This is a major release since the core was switched to use Qt 5 to get a more modern base platform, as well as to have an updated WebKit module.


freekmurze:
Hi, thank you for this.

Have you used the 64-bit binary that works on Ubuntu in this PR?

jamesmills:
Hi,

Yes. I have your package installed in my local Homestead box. I replaced the 1.9.7 version in bin with 2.1.1 version and ran my screenshot script and everything worked as expected.

There is still some webkit issues with PhantomJS itself

> PhantomJS 2.1.1 uses and older version of Webkit so the web standards are behind but you'll be glad to know that they are currently updating the the latest webkit - http://bit.ly/2bKI7EU

But from the release notes of 2.1 they seem to have improved some aspects of this.

I am still having a small issue with some of the websites I am taking a screen shot of but I am hoping that other webkit stuff will get fixed in future release.

James

freekmurze:
Cool, I'll tag a new version with the new binary.

Thanks again!

PR #31, opened 2016-11-11
setHeightToRenderWholePage() width issue

c2pdev:
Fix width issue when setHeightToRenderWholePage

### DIFF
--- a/src/Spatie/Browsershot/Browsershot.php
+++ b/src/Spatie/Browsershot/Browsershot.php
@@ -276,7 +276,7 @@ class Browsershot
             var page = require('webpage').create();
             page.settings.javascriptEnabled = true;
             page.settings.resourceTimeout = ".$this->timeout.';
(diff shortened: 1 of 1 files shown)


freekmurze:
The automated tests report syntax errors in this PR. Could you make sure the submitted code is correct and works?

c2pdev:
I've fixed the issue and tests are comming fine now

freekmurze:
Correct me if I'm wrong but it seems to me that this PR sets the height to 1 pixel when not specifying a height. Wouldn't this produce a image with a height of only one pixel?

c2pdev:
It may look weird but it works properly. If the height is set image will be rendered with  a proper value. If we don't set height it will use default value (480px). And most of all full width screenshot is rendered correctly.

c2pdev:
It looks like a bug in PhantomJS that ignores declared width if the height is  set to 0

freekmurze:
Ok, trusting you on this one!

Thanks for this PR

PR #36, opened 2017-02-02
Update allowed file extensions

al0mie:
There are possible cases when .png and .jpg not enough. Maybe worth allow more flexible settings for this like here http://phantomjs.org/api/webpage/method/render.html ?


freekmurze:
Thanks

PR #37, opened 2017-02-06
Generalize background feature

al0mie:
If phantom execute something like this, 
<pre>text = document.createTextNode('body { background: {$this->backgroundColor} }');</pre>
obviously it can set not only color and follow rule 
<pre>background: bg-color bg-image position/bg-size bg-repeat bg-origin bg-clip bg-attachment initial|inherit; </pre>
So maybe generalize this or use simple background-color ?


freekmurze:
Thank you for this but i'm not going to merge it in as most people will just want to pass a color name.

PR #42, opened 2017-03-10
Custom user agent

code2prog:
Added possibility to add a custom user agent because some firewalls may block the default one


freekmurze:
Please remove the `.idea` directory.

code2prog:
done

freekmurze:
Thanks!

PR #45, opened 2017-05-13
Added binary for mac

mihaliak:
This fixes `could not create screenshot` errors for MacOS users without using `setBinPath` and manually downloading phantomjs for Mac.


freekmurze:
Thanks

PR #62, opened 2017-07-03
Review

sebastiandedeyne:
- Why are the `paths` in `ChromeFinder` public?
- Can a custom Chrome path be used? Can't seem to find that


freekmurze:
- paths: I'll make them protected
- custom chrome path: there's a setChromePath for that

freekmurze:
Thanks!

PR #70, opened 2017-07-05
Escape arguments when creating the command

jdanino:


freekmurze:
Seems all good, thanks! 👍 

Don't mind the failing StyleCI check, I'll take care of that

PR #80, opened 2017-08-31
Add support for html input

itsgoingd:
Hey, this PR adds support for rendering arbitrary html input using a new `html` method:

```php
Browsershot::html('<h1>Hello world!!</h1>);
```

This is achieved by creating a temporary html file using the `TemporaryDirectory` component already used in the library, which is then passed to Chrome via `file://` scheme and cleaned up once we have the result.

Use case - a lot of rendered pdfs in web apps like invoices contain private information which makes it awkward to use with Browsershot, since you can't expose this data on a public url. An alternative is to use a temporary file, which can be easily done in a code wrapping Browsershot, but I think it's such a common case it would be nice to have this directly in Browsershot. This makes rendering an invoice as simple as:

```php
$pdf = Browsershot::html(view('invoice')->with(compact('order'))->render())
     ->export('invoice.pdf');
```

I've also added Chrome Canary path for MacOS and updated the test to pass with Canary.

Thanks for considering this feature!

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,12 @@ It will save a pdf if the path passed to the `save` method has a `pdf` extension
 Browsershot::url('https://example.com')->save('example.pdf');
 ```
 
+You can also use arbitrary html input, simply replace the `url` method with `html`:
+
+```php
+Browsershot::html('<h1>Hello world!!</h1>')->save('example.pdf');
+```
+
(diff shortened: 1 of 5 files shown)


freekmurze:
This is really nice. Thank you very much!

PR #81, opened 2017-09-19
Add high pixel density support

mikepostma:


freekmurze:
Nice, could you also add a test?

mikepostma:
Oops, yes. Overlooked the test.

freekmurze:
Perfect, thanks!

PR #83, opened 2017-09-27
Add the default path for linux Chromium users

aeyoll:
For instance, Debian users are most likely to have Chromium than Google Chrome on their system.


freekmurze:
Thanks!

PR #85, opened 2017-10-06
Rework to use Puppeteer instead of Chrome headless cli directly

itsgoingd:
Hey, this PR is a rework of the package internals using Puppeteer instead of the Chrome headless CLI directly, while keeping and expanding the public API.

[Puppeteer](https://github.com/GoogleChrome/puppeteer) is an official Node library from the Chrome team which provides an easy to use API for controlling Chrome over the DevTools protocol. This gives us access to a richer APIs allowing for a lot of customization options for the generated screenshots and pdfs that are not possible with the Chrome CLI.

### Pros

- way more customization options for generating screenshots and pdfs including some really useful ones
- not reinventing the wheel, using an official and actively developed 3rd party library
- not having to add and install Chrome from Google repositories (included in Puppeteer)

### Cons

- dependency on Node and Puppeteer (kinda like back in 1.x)
- breaking changes requiring a major release
- a slight performance hit (about 6s -> 8s for the original test suite)

### Screenshots

- *clip* - clipping a region of the page (might be redundant since we have the image manipulations api)
- *fullPage* - takes screenshot of the full scrollable page 👌 

### PDFs

- *browserHeaderAndFooter* - show/hide the browser header and footer
- *includeBackground* - include/exclude backgrounds
- *landscape* - landscape page orientation
- *pages* - printed page ranges, eg. `1-5, 8, 11-13`
- *pageSize* - page width/height in millimetres
- *margins* - page top/right/bottom/left margin in millimetres

### Removed APIs

I've removed these APIs since Puppeteer provides defaults in both cases.

- *disableGpu*/*enableGpu*
- *hideScrollbars*/*showScrollbars*

The [Puppeteer API](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md) provides a lot more options, not only for the screenshots and pdfs. I've decided to add only the most useful ones so the Browsershot API doesn't get out of hand.

Note that this can all be implemented in pure PHP since the DevTools protocol works over WebSockets. I've tried to hack on that solution, but it also required a large amount of composer dependencies and I didn't really like the code nor wanted to reinvent the wheel, so I went with Puppeteer.

This is a pretty big PR that has both pros and cons, so I have no expectations of you merging it. I've needed to write this anyway for my own use, so I might as well try to contribute back. 🙃 

Thanks for the consideration! 🚀


freekmurze:
What a great PR, thanks! I'll go over this soon.  No promises that I'm going to merge it, but if it all works and the code is clean enough I surely release a new major version to enable these features.

👍 

Could you already rebase against the current master to solve the conflicts?

itsgoingd:
Rebased, fixed code style and updated readme with new installation instructions API documentation. Not sure what to do about Travis though.

freekmurze:
Merging this already on a separate branch, so I can play with it and make some modifications to it.

I'll be sure to give you a shoutout in the credits section when this gets released.

Thanks again!

lukesnowden:
Just tested locally and does just what I was looking for, removing the header and footers! Thank you!

itsgoingd:
Nice, that’s pretty much the only reason I wrote this, trying to generate some invoices here. 😅

lukesnowden:
Yeah me too, cheers 🍻

freekmurze:
Currently swamped up in work / side projects, but I hope I have some time in the coming week to review this thoroughly and tag a new major version.

a3020:
> Currently swamped up in work / side projects, but I hope I have some time in the coming week to review this thoroughly and tag a new major version.

https://twitter.com/freekmurze/status/916924637380055040 😉

freekmurze:
😄

lukesnowden:
Had a few issues when getting it up and running for production on forge which I thought it may be worth mentioning for anyone else running into these issues;

1. You are no longer returning the process object from the `$this->callBrowser` method which the exception class is requiring (line 204/180).
2. Requires node upgrading to version `^8.0`
3. Puppeteer installing (obviously)
4. libi6 installing/updating by running `sudo apt-get install libxi6 libgconf-2-4`

regards

freekmurze:
@itsgoingd could you add the things mentioned by @lukesnowden to the readme?

itsgoingd:
New PR with improved error handling and installation instructions here https://github.com/spatie/browsershot/pull/87

barryvdh:
Nice!

But Puppeteer states that it only requires Node 6.4. Would it be possible to convert the browser.js script to support Node 6 (eg. compile with Babel to ES2016)

```
#!/usr/bin/env node
function _asyncToGenerator(fn) { return function () { var gen = fn.apply(this, arguments); return new Promise(function (resolve, reject) { function step(key, arg) { try { var info = gen[key](arg); var value = info.value; } catch (error) { reject(error); return; } if (info.done) { resolve(value); } else { return Promise.resolve(value).then(function (value) { step("next", value); }, function (err) { step("throw", err); }); } } return step("next"); }); }; }

const puppeteer = require('puppeteer');

const request = JSON.parse(process.argv[2]);

let fn = _asyncToGenerator(function* () {
    let browser, page;

    try {
        browser = yield puppeteer.launch();
        page = yield browser.newPage();

        if (request.options && request.options.userAgent) {
            yield page.setUserAgent(request.options.userAgent);
        }

        if (request.options && request.options.viewport) {
            yield page.setViewport(request.options.viewport);
        }

        yield page.goto(request.url);

        console.log((yield page[request.action](request.options)));

        yield browser.close();
    } catch (e) {
        if (browser) {
            yield browser.close();
        }

        console.error(e);
        process.exit(1);
    }
})();
```

I think that would probably make it easier to install, right?

freekmurze:
The Node 7 version of browser.js looks way cleaner. I'd leave it as it is. I don't have any problem with a requirement of Node 7.

itsgoingd:
Yeah, we can rewrite the script to use a generator or promises directly, but since Ubuntu 16.04 only has Node 4.2 and we need to use the official Node repository anyway, I figured we can go with the latest version. Also Node 8 should enter LTS later this month if we care about that stuff.

barryvdh:
Yeah obviously, this is the generated version from Babel ;)

But yeah, node 6.x is still current LTS at the moment, so suspect usage is still high and nodejs is always annoying to update, especially on shared servers (with version conflicts etc). And if the script can be rewritten cleanly, it shouldn't hurt much.

freekmurze:
I just want beautiful, modern code to maintain. 😄  Don't really care for people that can't upgrade their node version due to server restrictions.

PR #87, opened 2017-10-09
Puppeteer rework - improved error handling & installation instructions

itsgoingd:
Follow-up PR for https://github.com/spatie/browsershot/pull/85

- improved error handling when calling the browser
  - now correctly returns error instead of waiting for timeout on error
- fixed undefined variable usage & removed unused error handling code
  - I don't think we need the `CouldNotTakeBrowsershot` exception at all since Puppeteer should return error if it can't write the screenshot/pdf, but I left it in just to be sure
- updated installation instructions
  - based on clean Ubuntu 16.04 installation

We might consider installing Chrome via apt like before, so we can keep it easily updated and avoid installing all these dependencies explicitly - I can PR that later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,8 +55,9 @@ On a [Forge](https://forge.laravel.com) provisioned Ubuntu 16.04 server you can 
 
 ```bash
 curl -sL https://deb.nodesource.com/setup_8.x | sudo -E bash -
-sudo apt-get install -y nodejs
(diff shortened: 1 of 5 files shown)


freekmurze:
Thanks!

lukesnowden:
Might be worth noting the `travis-ci` is failing due to the node version, think this is set in the yaml file?

PR #89, opened 2017-10-10
sandbox option

sPooKee:
Now u can disable sandbox mode, when u'r running on a docker container.

### DIFF
--- a/src/Browsershot.php
+++ b/src/Browsershot.php
@@ -25,6 +25,8 @@ class Browsershot
     protected $userAgent = '';
     protected $deviceScaleFactor = 1;
 
+    protected $sandbox = true;
+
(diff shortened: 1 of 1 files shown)


freekmurze:
What is sandbox mode?

Could you also add a test and add an example of the new function to the readme?

sPooKee:
I tried to run this lib in a docker container and got the following error:

`Failed to move to new namespace: PID namespaces supported, Network namespace supported, but failed: errno = Operation not permitted`

After adding a `--no-sandbox` argument to the command, it worked.

I found this in Issue: https://github.com/jessfraz/dockerfiles/issues/65

freekmurze:
Could you add a test to prove that this works?

freekmurze:
Reopen when tests have been added.

PR #90, opened 2017-10-10
Page formats

TheNodi:
Puppeteer has a really handy features that allows you to specify paper format instead of sizes.

This makes for more readable and understandable code:
```php
// Paper size
Browsershot::url('https://example.com')->paperSize(297, 210)->save($targetPath);

// Paper format
Browsershot::url('https://example.com')->format('A4')->save($targetPath);
```

Available formats are listed here: https://github.com/GoogleChrome/puppeteer/blob/master/lib/Page.js#L780

We can also implement a "enum" class to list available formats instead of manually typing a string. 
Something as easy as:
```php
class PaperFormats
{
    const LETTER = 'letter';
    const LEGAL = 'legal';
    const TABLOID = 'tabloid';
    const LEDGER = 'ledger';
    const A0 = 'a0';
    const A1 = 'a1';
    const A2 = 'a2';
    const A3 = 'a3';
    const A4 = 'a4';
    const A5 = 'a5';
}

// Usage
Browsershot::url('https://example.com')->format(PaperFormats::A4)->save($targetPath);
```
This would make the class implementation-agnostic and user won't have to lookup puppeteer source code to find out if its format is available.

I can add a commit with this as well.

### DIFF
--- a/src/Browsershot.php
+++ b/src/Browsershot.php
@@ -27,6 +27,7 @@ class Browsershot
     protected $pages = '';
     protected $paperWidth = 0;
     protected $paperHeight = 0;
+    protected $format = null;
     protected $userAgent = '';
     protected $windowWidth = 800;
     protected $windowHeight = 600;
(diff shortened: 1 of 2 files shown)


freekmurze:
Thanks!

PR #94, opened 2017-10-17
Configureable node binary

brendt:
This pull requests adds the possibility to configure the node binary manually by calling the  `setNodeBinary` method.

Furthermore, the default environment in `browser.js` is removed. This fixed environment caused an error to occur when rendering a PDF within a web request. My best guess as to why this happens is because nginx doesn't use the same include paths as the terminal. The result was that PDFs could not be rendered on the fly in a Mac/nginx setup, because `node` was not a known environment.

By making the node binary configurable, this problem can be solved regardless of which system is running the library. A path to the node binary can be specified, and will default to `node`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,17 +65,18 @@ sudo npm install --global --unsafe-perm puppeteer
 sudo chmod -R o+rx /usr/lib/node_modules/puppeteer/.local-chromium
 ```
 
-### Custom node binary path
+### Custom node and npm binaries
 
-Depending on your setup, node might be not directly available to Browsershot. 
-If you need to manually set the node binary path, you can do this by calling the `setNodeBinary` method.
+Depending on your setup, node or npm might be not directly available to Browsershot. 
+If you need to manually set these binary paths, you can do this by calling the `setNodeBinary` and `setNpmBinary` method.
 
 ```
 Browsershot::html('Foo')
-    ->setNodeBinary('/usr/local/bin/node');
+    ->setNodeBinary('/usr/local/bin/node')
+    ->setNodeBinary('/usr/local/bin/npm');
 ``` 
 
-By default, Browsershot will use `node` to execute commands.
+By default, Browsershot will use `node` and `npm` to execute commands.
 
 ## Installation
 
--- a/README.md
+++ b/README.md
@@ -65,6 +65,18 @@ sudo npm install --global --unsafe-perm puppeteer
 sudo chmod -R o+rx /usr/lib/node_modules/puppeteer/.local-chromium
 ```
 
+### Custom node binary path
+
+Depending on your setup, node might be not directly available to Browsershot. 
+If you need to manually set the node binary path, you can do this by calling the `setNodeBinary` method.
+
+```
+Browsershot::html('Foo')
+    ->setNodeBinary('/usr/local/bin/node');
+``` 
+
+By default, Browsershot will use `node` to execute commands.
+
 ## Installation
 
 This package can be installed through Composer.
--- a/README.md
+++ b/README.md
@@ -78,6 +78,18 @@ Browsershot::html('Foo')
 
 By default, Browsershot will use `node` and `npm` to execute commands.
 
+### Custom include path
+
+If you don't want to manually specify binary paths, but rather modify the include path in general,
+you can set it using the `setIncludePath` method.
+
+```php
+Browsershot::html('Foo')
+    ->setIncludePath('$PATH:/usr/local/bin')
+```
+
+Setting the include path can be useful in cases where `node` and `npm` can not be found automatically.
+
 ## Installation
 
 This package can be installed through Composer.
--- a/bin/browser.js
+++ b/bin/browser.js
@@ -1,5 +1,3 @@
-#!/usr/bin/env node
-
 const puppeteer = require('puppeteer')
 
 const request = JSON.parse(process.argv[2])
--- a/src/Browsershot.php
+++ b/src/Browsershot.php
@@ -13,6 +13,7 @@ use Symfony\Component\Process\Exception\ProcessFailedException;
 class Browsershot
 {
     protected $nodeBinary = 'node';
+    protected $npmBinary = 'npm';
     protected $clip = null;
     protected $deviceScaleFactor = 1;
     protected $format = null;
@@ -55,6 +56,13 @@ class Browsershot
     public function setNodeBinary(string $nodeBinary)
     {
         $this->nodeBinary = $nodeBinary;
+
+        return $this;
+    }
+
+    public function setNpmBinary(string $npmBinary)
+    {
+        $this->npmBinary = $npmBinary;
 
         return $this;
     }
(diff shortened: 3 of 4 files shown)


freekmurze:
Waiting a bit on comments from @itsgoingd before merging this in.

itsgoingd:
Doesn't seem to work for me, the puppeteer module can't be found if I specify the node path.

The problem seems to be that `npm root -g` returns empty output if PATH is not correctly set, think they are calling node or using PATH internally.

Maybe it's just my environment tho, does it work for you?

An alternative solution to setting custom node path might be taking the `dirname` of the provided path and appending it to `PATH` in the CLI command. I'd also suggest appending common paths like `/usr/local/bin` which would make the package work out of the box on Mac and most environments.

brendt:
@itsgoingd Did you also test without specifying a custom binary? The default is set to `node`, which in almost all cases should work out of the box.

For my setup, this change does fix the issue. Did you install puppeteer globally or only for this project?

itsgoingd:
With or without specifying the node path the call ends with following error output:

```
env: node: No such file or directory
module.js:529
throw err;
^

Error: Cannot find module 'puppeteer'
at Function.Module._resolveFilename (module.js:527:15)
at Function.Module._load (module.js:476:23)
at Module.require (module.js:568:17)
at require (internal/module.js:11:18)
at Object.<anonymous> (/Users/its/Projects/browsershot-app/vendor/spatie/browsershot/bin/browser.js:1:81)
at Module._compile (module.js:624:30)
at Object.Module._extensions..js (module.js:635:10)
at Module.load (module.js:545:32)
at tryModuleLoad (module.js:508:12)
at Function.Module._load (module.js:500:3)
```

This is on latest macOS with Laravel Valet, latest node installed via brew and puppeteer installed globally via npm. Note that the original 3.0 release works just fine on Ubuntu and test suite runs on macOS.

brendt:
@itsgoingd just to be sure, could you share the contents of `vendor/spatie/browsershot/bin/browser.js`?

itsgoingd:
@brendt it is the latest version from the fix branch, I believe the node command not found error comes from the npm call which runs node via `/usr/bin/env node`.

Btw. this also works via `php artisan serve`, the only issue is in combination with Valet.

I made a sample project for testing, maybe we can get more Valet users to try it out? https://github.com/itsgoingd/browsershot-valet-test

brendt:
@itsgoingd Thanks for the test setup! I encountered the same problem, but after installing puppeteer locally, it did work.

```
npm install puppeteer
```

So it seems that using a global install of puppeteer + valet does not seem to work.

itsgoingd:
Yeah that works. I think what's happening is the `npm root -g` command still fails and returns no output, so we don't know the global node_modules path, but node falls back to using the local node_modules.

brendt:
An easy fix (as commited) is to also provide the option to set the npm path. This way we use the node binary to run the npm binary; which causes no errors.

To re-iterate: when using Browsershot in combination with Valet's webserver, your system's `PATH` is not set as expected. This causes `/usr/bin/env node` to fail on Mac, because these binaries live in `usr/local/bin`. It's however not as easy to fix this, because Mac doesn't allow you to make symlinks in `/usr/bin`. 

This PR adds the possibility to set both the `node` and `npm` binary paths, which make it work (because you're no relying on `PATH` anymore to determine where the binaries are located).

Another option however would be like in @itsgoingd 's PR: #93 , by manually resetting the `PATH` before running any commands. His solution doesn't require any additional setup by the programmers using this package.

This solution however offers a bit more flexibility to configure Browsershot on the fly, without the need of correctly setting path variables, like so:

```php
Browsershot::html('Foo')
    ->setNodeBinary('/usr/local/bin/node')
    ->setNodeBinary('/usr/local/bin/npm');
```

@freekmurze what's your opinion on which solution to go to?

If it's the second, there might be room for improvement. Maybe we could also add static setters so you wouldn't need to set the binaries every time you're using Browsershot? That's me thinking out loud here.

brendt:
I've also added @itsgoingd's original idea as a configurable property.

```php
Browsershot::html('Foo')
    ->setIncludePath('$PATH:/usr/local/bin')
```

freekmurze:
Seems ok to me, any thoughts before we tag this @itsgoingd ?

itsgoingd:
Looks fine to me! 👍

itsgoingd:
@brendt @freekmurze Hey, I believe this PR broke ootb support for global puppeteer installs, for a global install to work, the NPM path now **must** be explicitly set.

This is the generated command with default settings:

```
PATH=$PATH:/usr/local/bin NODE_PATH=`node npm root -g` node '/Users/its/Projects/oss/browsershot-test/vendor/spatie/browsershot/src/../bin/browser.js' '...'
```

The problematic part is the resolving of global node path - `node npm root -g` - since npm is now explicitly executed via node, it will be passed to node as-is without being resolved via PATH, this means node will try to execute script `npm` in CWD:

```
Error: Cannot find module '/Users/its/Projects/oss/browsershot-test/public/npm'
```

Now this also happens with local puppeteer install, the `node npm root -g` command fails, but in this case it doesn't matter since we are using the local path anyway.

A quick fix would be calling npm directly unless the npm path is explicitly set.

brendt:
@itsgoingd You mean that because the command is called as `node npm` it fails right? So if I understand this correctly: 

If a custom node path is set it should be added to execute npm, otherwise it should just be npm. So that would be something like this:

```php
if ($this->nodeBinary) {
    $setNodePathCommand = "NODE_PATH=`{$this->nodeBinary} {$this->npmBinary} root -g`";
} else {
    $setNodePathCommand = "NODE_PATH=`{$this->npmBinary} root -g`";
}
```

freekmurze:
Couldn't we just change the default value of `nodeBinary` from `node` to ``?

itsgoingd:
@brendt Yeah, for the `node $npmBinary` to work the `$npmBinary` has to be absolute path, so the check should look like this (assuming we consider all user-set npm paths to be absolute):

```php
if ($this->npmBinary === 'npm') {
    $setNodePathCommand = "NODE_PATH=`{$this->npmBinary} root -g`";
} else {
    $setNodePathCommand = "NODE_PATH=`{$this->nodeBinary} {$this->npmBinary} root -g`";
}
```

@freekmurze We still need to call node later in the CLI, but I still like the idea of taking dirname of the provided node & npm paths, prepending them to PATH and leave the work of finding them to OS, think that would simplify the code a bit.

brendt:
I was also thinking about `$nodeBinary` and `$npmBinary` being null by default. The real default value would be set in the method:

```php
$nodeBinary = $this->nodeBinary ?: 'node';

if ($this->npmBinary) {
    $setNodePathCommand = "NODE_PATH=`{$nodeBinary} {$this->npmBinary} root -g`";
} else {
    $setNodePathCommand = "NODE_PATH=`npm root -g`";
}
```

@itsgoingd If i'm correct, the above checks should solve the issue. 

@freekmurze Are you ok with the default approach?

itsgoingd:
@brendt Yep, that looks fine.

brendt:
PR was opened here: #108

PR #96, opened 2017-10-18
Add network idle timeout configuration

brendt:
This PR adds a configurable option to Browsershot with which you can set a network idle timeout. This parameter is passed directly to Puppeteer, and used to wait x-amount of request idle time before saving the page. This can be useful to wait for asynchronous fonts or images to load.

```
Browsershot::html('Foo')
    ->setNetworkIdleTimeout(100)
    ->save($targetPath);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -280,6 +280,16 @@ Browsershot::url('https://example.com')
     ->save($pathToImage);
 ```
 
+#### Set a network idle timeout
+
+You can tell Browsershot to wait x-amount of milliseconds before saving the HTML page.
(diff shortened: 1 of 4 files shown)


jnbn:
I think this should have a default value as setOption needs $value as a second argument that leads us to

```
Type error: Too few arguments to function Spatie\Browsershot\Browsershot::setOption(), 1 passed in /Users/jnbn/Sites/e
  pigverse/vendor/spatie/browsershot/src/Browsershot.php on line 109 and exactly 2 expected
```

I'll send a PR if that's OK for you too @freekmurze

PR #103, opened 2017-10-24
Added "no sandbox" option

itsgoingd:
Hey, this PR adds a `noSandbox()` option, which is needed when running linux in certain virtualization environments.


freekmurze:
Perfect, thanks!

PR #104, opened 2017-10-25
Fix "margin" option for PDF creation

martinfobian:
The browser.js option is actually called "margin" not "margins".


freekmurze:
Thank you!

PR #108, opened 2017-10-25
Default npm path bugfix

brendt:
This PR changes the way the `NODE_PATH` is set, so that npm can be run when no specific binary is supplied.

Only when a specific npm binary is supplied will the `NODE_PATH` be set using the custom paths. Otherwise, npm will be executed like normal.

### DIFF
--- a/src/Browsershot.php
+++ b/src/Browsershot.php
@@ -12,8 +12,8 @@ use Symfony\Component\Process\Exception\ProcessFailedException;
 /** @mixin \Spatie\Image\Manipulations */
 class Browsershot
 {
-    protected $nodeBinary = 'node';
-    protected $npmBinary = 'npm';
+    protected $nodeBinary = null;
+    protected $npmBinary = null;
     protected $includePath = '$PATH:/usr/local/bin';
     protected $networkIdleTimeout = 0;
     protected $clip = null;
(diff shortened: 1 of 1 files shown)


src/Browsershot.php
```diff
@@ -388,14 +388,20 @@ protected function callBrowser(array $command)
     {
         $setIncludePathCommand = "PATH={$this->includePath}";
 
-        $setNodePathCommand = "NODE_PATH=`{$this->nodeBinary} {$this->npmBinary} root -g`";
+        $nodeBinary = $this->nodeBinary ?: 'node';
+
+        if ($this->npmBinary) {
+            $setNodePathCommand = "NODE_PATH=`{$nodeBinary} {$this->npmBinary} root -g`";
+        } else {
```
freekmurze: Let's avoid this `else` statement by introducing a new `getNodePathCommand` function. Use early returns in that new function.

brendt:
@freekmurze I've added the `getNodePathCommand` method.

freekmurze:
Seems ok. Any remarks @itsgoingd ?

itsgoingd:
👍 for me

freekmurze:
Here we go! Thanks @brendt

freekmurze:
Here we go! Thanks @brendt

PR #111, opened 2017-10-25
ignoreHTTPSErrors by default

alexpgates:
This adds `ignoreHTTPSErrors: true` to the package by default in order to prevent https errors when adding non-https elements to a render.

### DIFF
--- a/bin/browser.js
+++ b/bin/browser.js
@@ -7,7 +7,10 @@ const callChrome = async () => {
     let page;
 
     try {
-        browser = await puppeteer.launch({ args: request.options.args || [] });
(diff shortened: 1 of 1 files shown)


alexpgates:
@freekmurze I wasn't sure how to structure a test with this pull request. I'm a bit inexperienced in this area, but I'm happy to dig in if it is necessary. No hard feelings if this PR won't work. I appreciate the opportunity to contribute!

freekmurze:
This is a breaking change: it changes the behaviour that people might rely upon now. 

I'd accept a PR that adds a function that sets that option.

PR #112, opened 2017-10-27
option to ignore HTTPs errors

alexpgates:
Here's another attempt at fixing HTTPS errors that may pop up.
Now you can pass an optional `->ignoreHttpsErrors()`
By default, HTTPS errors are not ignored.

I'm pretty new to this, so I'll add some comments to the PR so you can follow my thought process. Again, no hard feelings if this isn't what you're looking for. I really appreciate the opportunity to contribute, and I'm thankful for all your hard work. I'm open to any and all feedback / suggestions / etc.

Thank you!

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,14 +67,14 @@ sudo chmod -R o+rx /usr/lib/node_modules/puppeteer/.local-chromium
 
 ### Custom node and npm binaries
 
-Depending on your setup, node or npm might be not directly available to Browsershot. 
+Depending on your setup, node or npm might be not directly available to Browsershot.
 If you need to manually set these binary paths, you can do this by calling the `setNodeBinary` and `setNpmBinary` method.
 
 ```
 Browsershot::html('Foo')
     ->setNodeBinary('/usr/local/bin/node')
     ->setNpmBinary('/usr/local/bin/npm');
-``` 
+```
 
 By default, Browsershot will use `node` and `npm` to execute commands.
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -301,6 +301,16 @@ Browsershot::url('https://example.com')
...
+#### Ignore HTTPS errors
+
+You can ignore HTTPS errors, if necessary.
+
+```php
+Browsershot::url('https://example.com')
+   ->ignoreHttpsErrors()
+   ...
+```
+
```
alexpgates: Small addition to the README. I'm not sure how to best explain what causes the issue, so this may need some tweaking.
Also, I use `trim_trailing_white_space_on_save` which, I believe, explains why some of the other lines were modified.

bin/browser.js
```diff
@@ -7,7 +7,10 @@ const callChrome = async () => {
     let page;
 
     try {
-        browser = await puppeteer.launch({ args: request.options.args || [] });
+        browser = await puppeteer.launch({
+            ignoreHTTPSErrors: request.options.ignoreHttpsErrors,
```
alexpgates: To be honest, I wasn't sure if I should use a generic options array that gets merged into the config options array, or if it's okay to add the single option to the one line. There aren't many additional options (beyond the args) so I thought this was more readable. I'm interested if you have a strong opinion either way.

tests/BrowsershotTest.php
```diff
@@ -301,4 +301,25 @@ public function it_can_run_without_sandbox()
...
+            'options' => [
+                'ignoreHttpsErrors' => true,
+                'path' => 'screenshot.png',
+                'viewport' => [
+                    'width' => 800,
+                    'height' => 600,
+                ],
+            ],
+        ], $command);
+    }
```
alexpgates: I don't have much testing experience, but I'd like to get better. If this isn't what you have in mind for a test, please let me know.

freekmurze:
Seems good. Thanks 👍

PR #115, opened 2017-10-30
Allow the use of the 'omitBackground' option when capturing screenshots

rydurham:
This PR uses the `showBackground` value in the Browsershot class to dictate whether or not Puppeteer's `omitBackground` option should be set when capturing screenshots. 

This could potentially be seen as a breaking change because the `showBackground` value defaults to false, which means that this flag might be set for users without them wanting it to be there. 

I will send a separate PR for the `setOption` idea. 

Thanks!
~Ryan

### DIFF
--- a/src/Browsershot.php
+++ b/src/Browsershot.php
@@ -29,6 +29,7 @@ class Browsershot
     protected $paperHeight = 0;
     protected $paperWidth = 0;
     protected $showBackground = false;
+    protected $omitScreenshotBackground = false;
     protected $showBrowserHeaderAndFooter = false;
     protected $temporaryHtmlDirectory;
     protected $timeout = 60;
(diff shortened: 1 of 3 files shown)


freekmurze:
The background is currently visible when creating a screenshot.

Am I right in thinking that this PR will hide backgrounds for screenshots by default?

rydurham:
That is correct.   However, I think I can come up with a version that does not do that - I will update the PR.

rydurham:
Ok - I have updated the PR so that the default expectations are not changed.

src/Browsershot.php
```diff
@@ -29,6 +29,7 @@ class Browsershot
     protected $paperHeight = 0;
     protected $paperWidth = 0;
     protected $showBackground = false;
+    protected $omitScreenshotBackground = false;
```
freekmurze: Rename this to `showScreenshotBackground` make it `true`by default. Change up the added code to make it work that way.

That'll be more readable.

rydurham:
My commit message isn't that clear, but I believe I have now made the changes you were looking for.

freekmurze:
Thanks!

PR #118, opened 2017-11-06
Proxy Server Support

ngtech:


ngtech:
I am not sure however how to write a meaningful test for this without hardcoding a proxy server address / port that could later on change.

src/Browsershot.php
```diff
@@ -411,6 +419,14 @@ protected function createCommand(string $url, string $action, array $options = [
             $command['options']['args'] = ['--no-sandbox'];
         }
 
+        if ($this->proxy) {
+            if (isset($command['options']['args'])) {
+                $command['options']['args'][] = ['--proxy-server='.$this->proxy];
+            } else {
```
freekmurze: Could you avoid using an else statement? Use a helper function and early returns.

ngtech: I updated it to use a helper function to build the args array in the latest revision. The if part was incorrect anyway as it added an array instead of a string to the opts array.

src/Browsershot.php
```diff
@@ -407,8 +430,8 @@ protected function createCommand(string $url, string $action, array $options = [
             $command['options']['ignoreHttpsErrors'] = $this->ignoreHttpsErrors;
         }
 
-        if ($this->noSandbox) {
-            $command['options']['args'] = ['--no-sandbox'];
+        if ($args = $this->getOptionArgs()) {
```
freekmurze: Is this `if` necessary? `getOptionArray now always returns an array.

src/Browsershot.php
```diff
@@ -374,6 +382,21 @@ public function createPdfCommand($targetPath): array
         return $command;
     }
 
+    protected function getOptionArgs()
```
freekmurze: Add an `array` return type hint here.

src/Browsershot.php
ngtech: No it can be safely omitted - it is equivalent to what happens in line 12 of bin/browser.js
```javascript
  args: request.options.args || []
```

freekmurze: Remove it :-)

freekmurze:
I'll fix the nitpicks myself.

Thank you for this!

webaddicto:
Thanks for adding support for proxy, it is very useful!

One thing to notice is that it doesn't support socks5 authentication since Puppeteer doesn't support it yet:
[https://github.com/puppeteer/puppeteer/issues/4170](https://github.com/puppeteer/puppeteer/issues/4170)

As a possible workaround an user has wrote a module that supports it, info here:
[https://github.com/Cuadrix/puppeteer-page-proxy](https://github.com/Cuadrix/puppeteer-page-proxy)

This is just an information that can be useful to users that are looking for socks5 authentication.

Another possible option would be to set:

```
// Remember to base64 encode your username:password!
await page.setExtraHTTPHeaders({
    'Proxy-Authorization': 'Basic username:password',
    // OR
    Authorization: 'Basic username:password',
});
```

I will try to dig more and see if I can come up with a PR.

restyler:
I would like to note that https://github.com/Cuadrix/puppeteer-page-proxy does not use real browser when performing network requests: https://pixeljets.com/blog/how-to-set-proxy-in-puppeteer/#important-thing-to-know-about-github-packages-for-proxying-puppeteer-requests

PR #121, opened 2017-11-10
Move snapshot assertions to --dev

barryvdh:
This isn't actually required, is it? 

Right now it's installed always, even on production, which doesn't seem to be needed, or am I overlooking something?


freekmurze:
This is an oversight from our part. Thanks for this fix!

PR #129, opened 2017-11-13
Add the ability to allow arbitrary options to be passed to puppeteer

rydurham:
This PR adds a `setOption($key, $value)` method which will allow users to pass options to Puppeteer that are not explicitly covered by the rest of this package.   

Options can be set using dot notation, like so: 

```php
            ->showBackground()
            ->landscape()
            ->setOption('margin.top', '2cm')
            ->pages('1-3')
            ->paperSize(210, 148)
            ->createPdfCommand('screenshot.pdf');
```

### DIFF
--- a/src/Browsershot.php
+++ b/src/Browsershot.php
@@ -481,7 +481,7 @@ class Browsershot
 
         $command['options']['args'] = $this->getOptionArgs();
 
-        if (!empty($this->additionalOptions)) {
+        if (! empty($this->additionalOptions)) {
             $command['options'] = array_merge($command['options'], $this->additionalOptions);
         }
 
--- a/src/Browsershot.php
+++ b/src/Browsershot.php
@@ -41,6 +41,7 @@ class Browsershot
     protected $mobile = false;
     protected $touch = false;
     protected $dismissDialogs = false;
+    protected $additionalOptions = [];
 
     /** @var \Spatie\Image\Manipulations */
     protected $imageManipulations;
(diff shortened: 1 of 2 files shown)


src/Browsershot.php
```diff
@@ -274,6 +275,32 @@ public function windowSize(int $width, int $height)
...
 
+    public function setOption($key, $value)
+    {
+        if (is_null($key)) {
+            return $this;
+        }
+
+        $keys = explode('.', $key);
+
+        $array = &$this->additionalOptions;
```
freekmurze: Please refactor so there's no need for `&` in the code.

src/Browsershot.php
```diff
@@ -274,6 +275,32 @@ public function windowSize(int $width, int $height)
...
+
+        while (count($keys) > 1) {
+            $key = array_shift($keys);
+
+            if (! isset($array[$key]) || ! is_array($array[$key])) {
```
freekmurze: Use a null coalescing operator here.

tests/BrowsershotTest.php
```diff
@@ -390,4 +390,54 @@ public function it_can_use_a_proxy_server()
...
+                'foo' => [
+                    'bar' => 100,
+                    'baz' => 200,
+                ],
+            ],
+        ], $command);
+    }
+
+    /** @test */
+    public function it_can_set_arbitrary_options()
```
freekmurze: You used the same method name twice, causing the tests to fail.

src/Browsershot.php
```diff
@@ -454,6 +481,14 @@ protected function createCommand(string $url, string $action, array $options = [
 
         $command['options']['args'] = $this->getOptionArgs();
 
+        if (! empty($this->additionalOptions)) {
+            $command['options'] = array_merge($command['options'], $this->additionalOptions);
+        }
+
+        if (! empty($this->additionalOptions)) {
```
freekmurze: This seems like a duplication of the lines above.

rydurham:
Looks like I had some git rebase trouble there - sorry about that. 

Each of those requests should now be addressed.  Thanks!

freekmurze:
Could you also add some examples of how to use this method to the readme?

rydurham:
Sure thing.  Thanks!

rydurham:
It looks like you already took care of that - thanks!  I can still send another PR if you want.

freekmurze:
Nah, it's ok, thanks!

I've also refactor some of the internal to make use of the new `setOption` method.

Issue #131, opened 2017-11-16
Arbitrary filenames (without ext) result in MIME error


Synchro:
I've noticed that you can't use arbitrary filenames, even if you tell Browsershot what format to use. For example:

    Browsershot::url($url)
        ...
        ->format(Manipulations::FORMAT_JPG)
        ->save('image');

Results in:

```
The command "PATH=$PATH:/usr/local/bin NODE_PATH=`/Users/marcus/.nvm/versions/node/v9.1.0/bin/node /Users/marcus/.nvm/versions/node/v9.1.0/bin/npm root -g` /Users/marcus/.nvm/versions/node/v9.1.0/bin/node '/Users/marcus/Sites/site/admin/preview/vendor/spatie/browsershot/src/../bin/browser.js' '{"url":"https:\/\/www.example.net\/login.php","action":"screenshot","options":{"path":"temp","viewport":{"width":1280,"height":1024},"args":[],"fullPage":true,"dismissDialogs":true}}'" failed.

Exit Code: 1(General error)

Working directory: /Users/marcus/Sites/site/admin/preview

Output:
================


Error Output:
================
{ AssertionError [ERR_ASSERTION]: Unsupported screenshot mime type: application/octet-stream
    at Console.assert (console.js:193:23)
    at Page.screenshot (/Users/marcus/.nvm/versions/node/v9.1.0/lib/node_modules/puppeteer/lib/Page.js:632:15)
    at callChrome (/Users/marcus/Sites/site/admin/preview/vendor/spatie/browsershot/bin/browser.js:40:47)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
  generatedMessage: false,
  name: 'AssertionError [ERR_ASSERTION]',
  code: 'ERR_ASSERTION',
  actual: false,
  expected: true,
  operator: '==' }
```

So, setting `format` to `Manipulations::FORMAT_JPG` does not set the corresponding MIME type that ends up being passed to puppeteer, and puppeteer apparently doesn't like that.

freekmurze:
Could you PR a failing test (and a fix if you know a good way to fix this)?

freekmurze:
Handled in #134

PR #134, opened 2017-11-16
Failing test for #131

Synchro:
I've added a failing test for saving to filenames without extensions


freekmurze:
I've decided that the best course of action is to handle that error more gracefully.

Issue #132, opened 2017-11-16
Using sys temp folder results in error


Synchro:
I want Browsershot to store its temp files in the system tmp folder, so I do this to generate a filename:

    $tmpfile = tempnam(sys_get_temp_dir(), 'thumb_') . '.jpg';

giving a value like `/private/var/folders/ht/xx6x8mk88xjfyz001s7nn2800000gn/T/thumb_JtxCh5.jpg` on macOS. This folder exists and is writable by the user the script is running as. I pass it to Browsershot like so:

    Browsershot::url($url)
        ...
        ->format(Manipulations::FORMAT_JPG)
        ->save($tmpfile);

But this results in an error:

    Impossible to create the root directory "/var/folders/ht/xx6x8mk88xjfyz001s7nn2800000gn/T/thumb_JtxCh5.jpg"

Notice that the `/private` prefix has disappeared; on macOS, `/var` is a symlink to `/private/var`, so these two paths point to the same place, so it shouldn't be a problem. However, despite this error, it creates an empty file:

     -rw-------  1 marcus  staff     0B Nov 16 11:39 /private/var/folders/ht/xx6x8mk88xjfyz001s7nn2800000gn/T/thumb_JtxCh5

Note that it's missing the `.jpg` suffix, which is mysterious given what I found in #131.

If I provide a simple path like `./image.jpg` it seems happy, however, there is definitely something strange going on here. The `Impossible to create the root directory` message comes from [inside the `league/flysystem` package](https://github.com/thephpleague/flysystem/blob/master/src/Adapter/Local.php#L89). It looks like it's trying (and failing) to create a folder rather than a file. I wonder if it's caused by the difference between the paths with and without the `/private` prefix.

freekmurze:
This seems like a permission issue that is not really related to Browsershot.

Synchro:
It's definitely not permissions, and it's definitely inside Browsershot - this works perfectly in the same script using the same path:

    file_put_contents($tempfile, 'a');

Result:

    -rw-r--r--  1 marcus  staff     1B Nov 16 13:12 /private/var/folders/ht/xx6x8mk88xjfyz001s7nn2800000gn/T/thumb_USHVoG.jpg

Synchro:
Thanks for reopening. I can see that flysystem resolves symlinks [here](https://github.com/thephpleague/flysystem/blob/master/src/Adapter/Local.php#L76), however, this should convert `/var` to `/private/var`, not the other way around.

[...]

PR #136, opened 2017-11-17
Failing test case for #132

Synchro:


freekmurze:
See #132

Synchro:
Wouldn't it be better to merge this to catch regressions?

freekmurze:
Good point!

freekmurze:
Merge conflicts... I just copied the tests manually to the testsuite. Thanks again for this!

Issue #130, opened 2017-11-14
Stay on the URL for 2-5 sec


IAmMukeshKumar:
I have an animated page it takes 2-3 sec to complete the animation. I want screenshot at the end of the animation.

olavski:
Hi,

You can try and update the code in browser.js
 like this:
```javascript
await page.goto(request.url, requestOptions);

await page.waitFor( 30000 ); 

console.log(await page[request.action](request.options));
```

This should wait 3 seconds before taking the screenshot.
See if that works. :)

WidgetsBurritos:
@olavski Based on your recommendation here I opened a PR that adds this feature in #139, if you would like to check it out.

freekmurze:
We'll handle this further in #139

Issue #124, opened 2017-11-12
Dealing with lazy-loaded images


olavski:
First of all, thanks for a great library!! I've been waiting for years to be able to screenshots from PHP!

I was having problems taking full-page screenshots from long pages that was lazy-loading images.
I found one solution that works in some cases, it scrolls to the bottom of the page to trigger the loading of the images, then waits.

```javascript
await page.goto(request.url, requestOptions);

console.log("Scroll and wait...");
page.evaluate(_ => {
     window.scrollBy(0, document.body.scrollHeight);             
});
await page.waitFor( 30000 ); 

console.log(await page[request.action](request.options));
```
Even with a wait-time of 30 seconds it still hadn't loaded all the images, but when increasing the wait-time I ran into other timeouts. 
Now it scrolls from the top of the page directly to the bottom of the page.
A better solution might be to scroll down the page in stages, waiting for a bit before scrolling further.

Pages I tested on was http://www.thetimes.co.uk/ and http://www.independent.co.uk/

This isn't a complete solution to the problem I was having, but I can imagine other people are also dealing with this issue so I thought I'd put it out there. Maybe someone will know how to make this work :)

freekmurze:
We've added a `setDelay` method that hopefully will help capturing screenshots on those sites.

PR #139, opened 2017-11-20
Add delayed snapshot support

WidgetsBurritos:
Based on comments in #124 and #130, this PR adds the ability to set an optional delay on screenshot capture. This is useful when waiting for javascript or lazy-loaded resources to load before capturing. 

The included unit test is basically a best attempt at unit testing that. With potential network issues it's hard to guarantee that is always gonna be correct, without mocking a network request... which we can do if we need to.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,6 +205,15 @@ Browsershot::url('https://example.com')
     ->save($pathToImage);
 ```
 
+#### Delayed screenshots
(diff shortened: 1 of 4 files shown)


WidgetsBurritos:
Eh...  I kinda botched the unit test there.  I'll fix it in a little bit along with those style errors.

WidgetsBurritos:
Okay. I believe everything should be good to go now.

freekmurze:
Thank you very much!

WidgetsBurritos:
@freekmurze You're welcome.

PR #142, opened 2017-12-06
Add a way to set emulated media type page options

viktorivanov:
Based on puppeteer's emulateMedia option this adds method to set it easily. Especially helpful when generating pdfs, because by default it is set to request 'print' CSS media type of pages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,6 +327,18 @@ Browsershot::url('https://example.com')
     ->save($pathToImage);
 ```
 
+#### Setting the CSS media type of the page
+
+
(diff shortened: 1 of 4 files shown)


viktorivanov:
@freekmurze can you have a look at this please :)

freekmurze:
Looks good, could you also add an example to the readme?

viktorivanov:
@freekmurze, did you had the time to check it now?

freekmurze:
Thanks!

Issue #140, opened 2017-11-29
Output image or PDF to browser


clementmas:
Is there any way to output the screenshot or PDF straight to the browser without having to create a new file?

Here's how I've been doing it with Node.js and Express:
```javascript
app.get('/screenshot', async (req, res) => 
{
    const browser = await puppeteer.launch();

    const page = await browser.newPage();
    await page.goto('https://example.com');

    let image = await page.screenshot();

    await browser.close();

    res.writeHead(200, {'Content-Type': 'image/png'});
    res.end(image);
});
```

freekmurze:
Currently there is no such method. I'd accept a PR for this.

freekmurze:
Added via #144

PR #144, opened 2017-12-06
Output image or PDF to browser #140

clementmas:
Added 2 new methods `screenshot()` and `pdf()` to output the screenshot or PDF straight to the browser without having to create a new file.

### DIFF
--- a/bin/browser.js
+++ b/bin/browser.js
@@ -5,6 +5,7 @@ const request = JSON.parse(process.argv[2]);
 const callChrome = async () => {
     let browser;
     let page;
+    let output;
 
     try {
         browser = await puppeteer.launch({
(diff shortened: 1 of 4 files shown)


bin/browser.js
```diff
@@ -41,7 +42,16 @@ const callChrome = async () => {
             await page.waitFor(request.options.delay);
         }
 
-        console.log(await page[request.action](request.options));
+        output = await page[request.action](request.options);
+
+        if ( request.options.path )
+        {
+            console.log(output);
```
freekmurze: Please remove all code regarding debugging.

src/Browsershot.php
```diff
@@ -289,6 +289,24 @@ public function bodyHtml(): string
         return $this->callBrowser($command);
     }
 
+    public function screenshot()
```
freekmurze: This returns a string right? Add a `string` return type hint

src/Browsershot.php
```diff
@@ -289,6 +289,24 @@ public function bodyHtml(): string
...
+    {
+        $command = $this->createScreenshotCommand();
+
+        $encoded_image = $this->callBrowser($command);
+
+        return base64_decode($encoded_image);
+    }
+
+    public function pdf()
```
freekmurze: Add a string return type hint

freekmurze:
Thanks!

PR #146, opened 2017-12-08
Add puppeteer chrome executable path support

WidgetsBurritos:
Per the [puppeteer api docs](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#puppeteerlaunchoptions), we can specify alternate executable path for chrome/chromium: 

>   - `executablePath` <[string]> Path to a Chromium executable to run instead of bundled Chromium. If `executablePath` is a relative path, then it is resolved relative to [current working directory](https://nodejs.org/api/process.html#process_process_cwd).

This PR adds support for this feature.

### DIFF
--- a/src/Browsershot.php
+++ b/src/Browsershot.php
@@ -86,6 +86,13 @@ class Browsershot
         return $this;
     }
 
+    public function setExecutablePath(string $executablePath)
+    {
+        $this->setOption('executablePath', $executablePath);
+
+        return $this;
+    }
+
     /**
      * @deprecated This option is no longer supported by modern versions of Puppeteer.
      */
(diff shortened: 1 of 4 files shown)


src/Browsershot.php
```diff
@@ -86,6 +86,13 @@ public function setNodeModulePath(string $nodeModulePath)
         return $this;
     }
 
+    public function setExecutablePath(string $executablePath)
```
freekmurze: Rename this to `setChromePath`

WidgetsBurritos:
@freekmurze Okay I pushed up a new commit with the new method name.

freekmurze:
Thanks!

PR #152, opened 2018-01-19
Feature: Custom Header and Footer HTML

AdrianB93:
Add custom header and footer HTML via `headerHtml($someHtml)` and `footerHtml($someHtml)`.

Also added `hideHeader()` and `hideFooter()`, but not `showHeader()` or `showFooter()` as `showHeader()->showFooter()` would have different behaviour to `showBrowserHeaderAndFooter()`.

Requires puppeteer 1.0.0.


freekmurze:
Thanks!!

PR #154, opened 2018-01-24
Added "wait until network idle" option

itsgoingd:
Hey, this PR adds a `waitUntilNetworkIdle()` option as a replacement for the old `setNetworkIdleTimeout()` option when using Puppeteer 1.x.

- `waitUntilNetworkIdle()` by default waits for the `networkidle0` event - no network connections for 500 ms
- `waitUntilNetworkIdle($strict = false)` waits for the `networkidle2` event - no more than 2 network connections for 500 ms (useful in edge cases where the website is pinging some ajax endpoint)

I believe this is in most cases better than using the delay option as with delay you are just guessing how long the resource loading will take, so you either waste time waiting or not waiting long enough. With network idle you are only waiting for an extra 500 ms in the best case.

Also I've found this required for webfonts to be properly rendered in some cases, not only SPAs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,6 +223,17 @@ Browsershot::url('https://example.com')
     ->save($pathToImage);
 ```
 
+#### Waiting for lazy-loaded resources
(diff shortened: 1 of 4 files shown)


freekmurze:
Nice 👍 

Could you update the readme with an description and example of the new function?

itsgoingd:
Added readme section. 👌

freekmurze:
Thanks!

PR #162, opened 2018-02-19
Add support for custom binary/browser script

retiolum:
I've added a new options to make usage of a custom binary/browser script so that custom adjustments can be made without touching the library code.


freekmurze:
Thanks!

Could you update the readme as well?

retiolum:
README is now amended.

freekmurze:
Thanks!

PR #165, opened 2018-03-04
Setting browser locale

nicolas-t:
Example of how to use `setOption` in order to change the language of a page.
Maybe we could do something more generic about puppeteer's `args` ?


freekmurze:
Thanks!

PR #172, opened 2018-03-27
Add support for taking screenshot of an element using a selector

dzava:
This PR adds a new method `select` which can be used to capture a screenshot of an element matching a selector. It works by finding the [boundingBox](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#elementhandleboundingbox) of the element and using it to clip the screenshot.

A coupe things to note:
- If the selector does not match an element, ~~it is ignored~~ an exception is thrown
- it has no effect when creating a pdf, since `clip` is not supported when generating pdf files

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,14 @@ Browsershot::url('https://example.com')
     ->save($pathToImage);
 ```
 
+You can take a screenshot of an element matching a selector using `select`.
+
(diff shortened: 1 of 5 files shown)


freekmurze:
Pretty awesome PR, thanks!

PR #173, opened 2018-03-28
Add extraHTTPHeaders options

### LINKED ISSUE #171: Send custom header
For our use case we need to send an authorization header with our request. I can't find a way to support this here. Is it an option?

artggd:
Hi,
This PR allows to add HTTP headers to the page call using [extraHTTPHeaders](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagesetextrahttpheadersheaders) option. Fixes #171

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to `Browsershot` will be documented in this file
 
+### 3.18.0 - 2018-03-28
+
+- add support for taking screenshot of an element using a selector
+
 ### 3.17.0 - 2018-02-22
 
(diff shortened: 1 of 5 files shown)


freekmurze:
Thanks! Could you also update the readme and add a test?

artggd:
I added a test and updated the readme 🙂

freekmurze:
Great, thanks!

ChrisWiegman:
Thanks for addressing #171 so quickly, particularly as it was more than just something I missed in the docs somewhere.

artggd:
All good mate, I needed this feature on my use case 🙂

Issue #177, opened 2018-04-11
Add support for various options on screenshot() function


darron1217:
https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagescreenshotoptions

Firstly, I really appreciate with this fancy package :)
Anyway, Puppeteer supports `type` option of screenshot image. (both png and jpeg)
It would be very nice if we have a way to pass various options like this `screenshot($options)`

PR #178, opened 2018-04-11
Add support for various config on screenshot()

darron1217:
Issue #177

### DIFF
--- a/src/Browsershot.php
+++ b/src/Browsershot.php
@@ -337,7 +337,7 @@ class Browsershot
             return $this->savePdf($targetPath);
         }
 
-        $command = $this->createScreenshotCommand($targetPath);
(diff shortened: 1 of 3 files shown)


src/Browsershot.php
```diff
@@ -404,15 +404,10 @@ public function createBodyHtmlCommand(): array
         return $this->createCommand($url, 'content');
     }
 
-    public function createScreenshotCommand($targetPath = null): array
+    public function createScreenshotCommand($options = []): array
```
freekmurze: As it stands this is a breaking change. This function should still be able to handle string input

darron1217: You mean that this needs backward compatibility?

darron1217:
I've created another PR #180 to solve this issue simply.

PR #179, opened 2018-04-11
Remove typehint from emulate media

fourstacks:
The docs suggest that you can set emulateMedia to null to remove emulation, however the emulateMedia method in browsershot is typehinted to a string so this option is not currently possible.  This PR fixes this by removing the typehint.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ addons:
 language: php
 
 php:
-  - 7.0
   - 7.1
   - 7.2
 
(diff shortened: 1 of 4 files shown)


fourstacks:
Hey @freekmurze.  Good suggestion - I wasn't aware that optional typehints were a thing so every day's a school day!  Updated the PR accordingly

freekmurze:
Nice, could you also set the minimum required version to 7.1 in `composer.json` and update `.travis.yml` so tests won't run using PHP 7.0?

fourstacks:
All done :)

freekmurze:
Great, thanks!

PR #180, opened 2018-04-12
Add setScreenshotType() method + Update Readme

darron1217:
Another simpler implementation of Issue #177

### DIFF
--- a/src/Browsershot.php
+++ b/src/Browsershot.php
@@ -223,7 +223,7 @@ class Browsershot
     public function setScreenshotType(string $type, int $quality = null)
     {
         $this->screenshotType = $type;
-        if(!is_null($quality)) {
+        if (! is_null($quality)) {
             $this->screenshotQuality = $quality;
         }
 
--- a/src/Browsershot.php
+++ b/src/Browsershot.php
@@ -422,7 +422,7 @@ class Browsershot
         $url = $this->html ? $this->createTemporaryHtmlFile() : $this->url;
 
         $options = [
-            'type' => $this->screenshotType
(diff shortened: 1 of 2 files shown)


src/Browsershot.php
```diff
@@ -23,6 +23,8 @@ class Browsershot
     protected $proxyServer = '';
     protected $showBackground = false;
     protected $showScreenshotBackground = true;
+    protected $screenshotType = '';
```
freekmurze: Set this to  `png` by default.

src/Browsershot.php
```diff
@@ -413,6 +425,14 @@ public function createScreenshotCommand($targetPath = null): array
             $options['path'] = $targetPath;
         }
 
+        if ($this->screenshotType) {
```
freekmurze: Remove this `if` statement.

src/Browsershot.php
```diff
@@ -23,6 +23,8 @@ class Browsershot
     protected $proxyServer = '';
     protected $showBackground = false;
     protected $showScreenshotBackground = true;
+    protected $screenshotType = '';
+    protected $screenshotQuality = 0;
```
freekmurze: Set this to `null` by default.

src/Browsershot.php
```diff
@@ -218,6 +220,16 @@ public function hideBackground()
         return $this;
     }
 
+    public function setScreenshotType(string $type, int $quality = null)
+    {
+        $this->screenshotType = $type;
```
freekmurze: Add an empty line after this one.

freekmurze:
Thanks!

PR #181, opened 2018-04-13
Add evaluate() function

darron1217:
I've created a PR to add puppeteer's `evaluate()` function
https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pageevaluatepagefunction-args

### DIFF
--- a/bin/browser.js
+++ b/bin/browser.js
@@ -2,20 +2,6 @@ const puppeteer = require('puppeteer');
 
 const request = JSON.parse(process.argv[2]);
 
-const getOutput = async (page, request) => {
-    let output;
-
-    if (request.action == 'evaluate') {
-        output = await page.evaluate(request.options.pageFunction);
-
-        return output;
-    }
-
-    output = await page[request.action](request.options);
-
-    return output.toString('base64');
-};
-
 const callChrome = async () => {
     let browser;
     let page;
@@ -87,7 +73,12 @@ const callChrome = async () => {
             request.options.clip = await element.boundingBox();
         }
 
-        output = await getOutput(page, request);
+        if(request.action == 'evaluate') {
+            output = await page.evaluate(request.options.pageFunction);
+        } else {
+            output = await page[request.action](request.options);
+            output = output.toString('base64')
+        }
 
         if (!request.options.path) {
             console.log(output);
--- a/bin/browser.js
+++ b/bin/browser.js
@@ -2,6 +2,15 @@ const puppeteer = require('puppeteer');
 
 const request = JSON.parse(process.argv[2]);
 
+const getOutput = (page, request) => {
+
+    if(request.action == 'evaluate') {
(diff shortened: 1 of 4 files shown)


darron1217:
I've also added tests error fix

tests/BrowsershotTest.php
```diff
@@ -162,20 +162,6 @@ public function it_can_handle_a_permissions_error()
             ->save($targetPath);
     }
 
-    /** @test */
-    public function it_can_use_the_methods_of_the_image_package()
```
darron1217: I removed this test because there's method named `format()` is overrided. So we cannot access to spatie/image package's function.
(I guess `format()` function is implemented for setting pdf format)

tests/BrowsershotTest.php
```diff
@@ -201,6 +187,7 @@ public function it_can_create_a_command_to_generate_a_screenshot()
                     'height' => 1080,
                 ],
                 'args' => [],
+                'type' => 'png',
```
freekmurze: Why is 'png' added in every test? Can this be removed?

bin/browser.js
```diff
@@ -73,10 +73,15 @@ const callChrome = async () => {
             request.options.clip = await element.boundingBox();
         }
 
-        output = await page[request.action](request.options);
+        if(request.action == 'evaluate') {
+            output = await page.evaluate(request.options.pageFunction);
+        } else {
```
freekmurze: Can you refactor this, so we don't need an `else` statement?

First thought: let's introduce a `getOutput` function. You can use an early return in that function.

tests/BrowsershotTest.php
darron1217: It related to https://github.com/spatie/browsershot/pull/180

This change is not related with this PR. Should I remove it?
(I just tried to remove Travis Error)

bin/browser.js
darron1217: Can you explain more about `getOutput`?

You mean `request.action == 'getOutput'`?

darron1217: I implemented it on d33189d31bd05bc2a26edc851ea20cbca7c40e0b
Is it the function you said?

darron1217:
I've implemented getOutput() on browser.js

freekmurze:
The tests seem to be failing. Can you also add a test for the  `evaluate` function?

darron1217:
The test error is caused by another accepted PR #180 
I'll create another PR to fix test error

freekmurze:
Thanks! Could you rebase to the current master?

darron1217:
Okay wait for a second

darron1217:
I rebased it, and commits are dirty... haha

darron1217:
I don't understand how to rebase it....
Can you help me with doing it?

darron1217:
#185
I created cleaned version of this PR

Issue #166, opened 2018-03-08
Navigation Timeout does not seem to be set by `->timeout()` option


tpavlek:
When using browshot I'm consistently getting navigation timeout exceeded 30,000ms - which is 30 seconds. Changing the timeout option does not seem to affect this in any way.

anthonyjdalfonso:
I second this. Would be as simple as setting the option in **Browsershot.php**

```php
public function timeout(int $timeout)
{
    $this->timeout = $timeout;
    $this->setOption('timeout', $timeout);

    return $this;
}
```

and then consuming it in **browser.js**
```javascript
if (request.options && request.options.timeout) {
    await page.setDefaultNavigationTimeout(request.options.timeout);
}
```

I can't think of a reason why Browsershot and Puppeteer would operate on separate timeouts.

PR #187, opened 2018-04-19
Set the timeout used by Puppeteer.

SudoGetBeer:
First of all this my first PR so be warned ;)

This PR would set the timeout used by Puppeteer.

Should fix issue:
#166


SudoGetBeer:
Thank you.
Yay first PR and it got accepted.

PR #190, opened 2018-04-20
Fix typo in test where it checked the same thing twice

TVke:
the test was there to test the wrong `setBinPath` and the `setChromePath` already existed


freekmurze:
Thanks!

Issue #189, opened 2018-04-20
For some reason Chrome did not write a file at 'location'


techtic-harshad:
In CouldNotTakeBrowsershot.php line 11:

Browsershot::url('https://google.com')->save(public_path().'/google.png');

i will try to implement this but received error, anyone please help me with that?

SudoGetBeer:
What was the error?

techtic-harshad:
it always through the error from CouldNotTakeBrowsershot.php and says like "For some reason Chrome did not write a file at 'location/path'"

ciprianoprea89:
Same problem here. (OS : windows 10)

freekmurze:
I'm not a windows user, and tbh don't keep windows compatibility in mind when creating stuff.

I'd accept a PR that fixes this and is verified to work on both windows, mac and linux systems.

shaheenkhan401:
Hey danielsdeboer,
I am using window 10. I have used addslashes() but still getting (For some reason Chrome did not write a file at `D:\xampp\htdocs\crase\public/a.png`.) is there any other solution for this ??

standus:
It doesnt work for me on my Windows 10. 
I see that in functon callBrowser is $this->includePath is command for linux. Then there is $binPath with /bin/browser.js but on Win platform browser.js is at another path: lib/bin/browser.js .

colinmackinlay:
Same issue for me on Win10 (Laragon). I can see the commit that addresses the above issue is in the latest release so something else is presumably stopping Chrome writing the files. Anyone found out why?

raana96:
Hi, I have the same problem. My operating system is Windows 10. I changed escapeshellarg() to addslashes() but again I got the same error.  Please guide me if you solve the problem. Thank you.

ndeblauw:
Installing puppeteer locally (thus: `npm install puppeteer`) instead of globally did the trick for me. 

The only thing that I find very strange is that **it works when ran from the command line** (console command, even in tinker) but **not when the command originates from a web request** (in a controller)...

colinmackinlay:
> Installing puppeteer locally (thus: `npm install puppeteer`) instead of globally did the trick for me [Win10, Laragon, PHP 7.4].
> 
> The only thing that I find very strange is that **it works when ran from the command line** (console command, even in tinker) but **not when the command originates from a web request** (in a controller)... The trick for getting it to work there (for me) is to queue a job that generates it. Queue workers do manage to generate the image through the job.

[...]

PR #199, opened 2018-06-06
Fix for #189 - no screenshot created on Windows systems

MrSnoozles:
This changes the command generation on Windows systems.


freekmurze:
Thanks!

PR #204, opened 2018-07-05
Add waitForFunction() method

darron1217:
There's nice waitForFunction on puppteer API

Let's utilize it :)

### DIFF
--- a/src/Browsershot.php
+++ b/src/Browsershot.php
@@ -139,7 +139,7 @@ class Browsershot
         return $this;
     }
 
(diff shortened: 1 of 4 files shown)


src/Browsershot.php
```diff
@@ -139,6 +139,14 @@ public function waitUntilNetworkIdle(bool $strict = true)
         return $this;
     }
 
+    public function waitForFunction($function, $polling = 'raf', $timeout = 0)
```
freekmurze: - Could you typehint all arguments?
- What is `raf`?

freekmurze:
Thanks!

Issue #214, opened 2018-08-10
POST result screenshot


delia-curiel:
Hi, 
what would be the best way to get a screenshot of a page, that needs a previous post by ajax to get the results?   I mean, I need the screenshot of the results after executing the search, that is done by a post request

Thanks!

freekmurze:
I don't think it possible now but I'd love a PR that adds this to the package.

API could be something like this:

```
Browsershot::url('https://example.com')->usePost([
   'name' => 'value',
])->save('example.pdf');
```

ctf0:
a workaround could be, add a js code that reads the query string from the url & populate the search field with the text, then click the search btn ex
```js
Browsershot::url('https://example.com?search=test')
    ->click('#search-btn')
    ->setDelay(500) // just to make sure we got some results b4 saving the ss
    ->save('example.pdf')
```

tomhatzer:
Puppeteer is not capable of doing post requests on commandline ( you have to intercept the request - have a look at https://github.com/GoogleChrome/puppeteer/issues/669 ), so the post request itself would have to happen in a separate call and the resulting HTML could be interpreted by puppeteer.
Does this sound like a sane way to implement the wanted functionality?
If so, which HTTP client library would be good for such a task? Guzzle? Plain PHP with file_get_contents() or sockets? Any preferences?

freekmurze:
If it's easy to do, try doing it with plain PHP.

tomhatzer:
Ok I tried to get this thing running. If anyone has some time and could look at my PR that would be great. Ideas/suggestions/improvements/fixes/... are welcome.

delia-curiel:
Thanks everyone for your answers... I'm not very sure if what I needed require changes in the package code....
I would need something similar to what @ctf0 have answered.... I mean, the sequence of calls needed to obtain the result:
1. access url  (which will be a page with a search box)
2. set a value on the search box
3. click the search button
4. screenshot the results

freekmurze:
There's already a PR for this: #216

@delia-curiel could you test out the code in that PR and let me know (in #216) if that works for you?

delia-curiel:
@freekmurze Hi again, I have tested with my code and it does not return the desired results....

Here is the site I want to get the screenshot from:
https://online.austrac.gov.au/ao/public/rsregister.seam

I want to get the screenshot with the results after searching a value.
I have use the code from the PR like that:

$url = "https://online.austrac.gov.au/ao/public/rsregister.seam";
[...]

PR #216, opened 2018-08-15
Add method to do post requests in browsershot

tomhatzer:
First version of a solution for issue https://github.com/spatie/browsershot/issues/214

### DIFF
--- a/tests/BrowsershotTest.php
+++ b/tests/BrowsershotTest.php
@@ -910,4 +910,24 @@ class BrowsershotTest extends TestCase
             ],
         ], $command);
     }
+
+    /** @test **/
+    public function it_can_send_post_parameters_and_get_the_body_html()
+    {
+        $html = Browsershot::url('http://example.com')
+            ->usePost(['hello' => 'hi there!', 'is it me' => 'you are looking for?'])
+            ->bodyHtml();
+
+        $this->assertContains('<h1>Example Domain</h1>', $html);
+    }
+
+    /** @test **/
+    public function it_can_send_post_parameters_to_ssl_and_get_the_body_html()
+    {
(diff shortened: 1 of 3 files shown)


tests/BrowsershotTest.php
```diff
@@ -910,4 +910,24 @@ public function it_can_add_a_wait_for_function_to_puppeteer()
...
     }
+
+    /** @test **/
+    public function it_can_send_post_parameters_and_get_the_body_html()
+    {
+        $html = Browsershot::url('http://example.com')
+            ->usePost(['hello' => 'hi there!', 'is it me' => 'you are looking for?'])
+            ->bodyHtml();
+
+        $this->assertContains('<h1>Example Domain</h1>', $html);
```
freekmurze: How does this output prove that `usePost` is actually working?

freekmurze:
Thank you for this, I'll do a thorough review soon.

tests/BrowsershotTest.php
tomhatzer: There's currently no way of proving this using example.com as domain. We'd have to use something else like httpbin.org or another http post service that returns the post values. If that's possible I'll customize the 2 tests.

freekmurze:
@delia-curiel reported in #214 that this code is not working. 

@tomhatzer could you take a look at this?

tomhatzer:
I'll check this again. Posting data works, but that website has some header checks or something else. I'll see what I can come up with. Will post here and update the PR as soon as I've got a working solution.

tomhatzer:
@freekmurze this PR is nonsense, please ignore it. I'm going to open a new one with working code... Sorry for causing trouble.

PR #220, opened 2018-09-01
Add type() method to js and php files to allow post requests

tomhatzer:
Add type() method to browser.js and Browsershot.php and also include tests for new method and update README.md

Working fix for https://github.com/spatie/browsershot/issues/214


freekmurze:
Thanks!

PR #227, opened 2018-10-09
Device emulate support

shahabphp:
Implements device emulate.
Devices list:
https://github.com/GoogleChrome/puppeteer/blob/07febb637c78cd59e22a15166f816d838a36e614/DeviceDescriptors.js#L17


freekmurze:
Thank you for this. Could you update the readme with a few good example on how to use the new functionality?

shahabphp:
I documented the `device` method and updated the readme file.

Issue #230, opened 2018-10-17
chrome processes are not closing


zarasargsyan:
Hello,

I'm using html to image functionality: _Browsershot save()_. 
There are timeout errors sometimes (I've changed timeout and it's ok), but I have noticed that when I'm getting an error - there are opened chrome processes by library on server side and they are not closing. 
I have found the solution too. 
In Browsershot class, callBrowser function should be $process->clearOutput(); at line 581

        if ($process->isSuccessful()) {
            return rtrim($process->getOutput());
        }
        $process->clearOutput();

        if ($process->getExitCode() === 2) {
            throw new ElementNotFound($this->additionalOptions['selector']);
        }
I can create a pull request

freekmurze:
Feel free to PR this! 👍

PR #232, opened 2018-10-18
Issue 230

zarasargsyan:
resolving not closed chrome processes issue #230

### DIFF
--- a/src/Browsershot.php
+++ b/src/Browsershot.php
@@ -578,7 +578,7 @@ class Browsershot
         if ($process->isSuccessful()) {
             return rtrim($process->getOutput());
         }
-		$process->clearOutput();
+		 $process->clearOutput();
 
         if ($process->getExitCode() === 2) {
             throw new ElementNotFound($this->additionalOptions['selector']);
--- a/src/Browsershot.php
+++ b/src/Browsershot.php
@@ -578,6 +578,7 @@ class Browsershot
         if ($process->isSuccessful()) {
             return rtrim($process->getOutput());
         }
+		$process->clearOutput();
 
         if ($process->getExitCode() === 2) {
             throw new ElementNotFound($this->additionalOptions['selector']);


hashcrypt-devs-rajkot:
What if process is success ? Then it will also not close the process because it will return the output from Line 623 and "$process->clearOutput()" is at Line 627.

andi-wr:
Same problem here. How is it intended that the chrome process is closed - when the php process is finished ("closed")?

I use browsershot inside an event loop of ReactPHP, so the "wrapping" php process will run forever because it acts as a server by itself.

Now I have hundrets of chrome processes looking like:

```
...
root     31748  0.0  0.0      0     0 ?        Z    Jul15   0:00 [chrome] <defunct>
root     31765  0.0  0.0      0     0 ?        Z    Jul15   0:00 [chrome] <defunct>
root     31767  0.0  0.0      0     0 ?        Z    Jul15   0:00 [chrome] <defunct>
root     31781  0.0  0.0      0     0 ?        Z    Jul15   0:00 [chrome] <defunct>
```

Minimum code extract:

```php
<?php
include "vendor/autoload.php";

use Spatie\Browsershot\Browsershot;
use React\EventLoop\Factory as EventLoopFactory;

const LISTEN_PORT = 2380;
const LISTEN_IP = "0.0.0.0";

const DIMENSION_UNITS = ['px', 'in', 'cm', 'mm'];
const EMULATE_MEDIA_VALUES = ['screen', 'print'];

$loop = EventLoopFactory::create();

$server = new React\Http\Server(function (Psr\Http\Message\ServerRequestInterface $request) {
    $getParams = $request->getQueryParams();
    $url = $getParams['url'];

    echo "Incoming Request for URL \"" . $url . "\"\n";

    $browsershot = new Browsershot($url);

    // general options
    $browsershot->noSandbox()->fullPage()->waitUntilNetworkIdle();

    // ... do some stuff - set options etc. ...

    $out = $browsershot->pdf();

    return new React\Http\Response(
        200,
        ['Content-Type' => 'application/pdf'],
        $out
    );
});

$socket = new React\Socket\Server('tcp://'.LISTEN_IP.':' . LISTEN_PORT, $loop);
$server->listen($socket);

echo "Server running at http://".LISTEN_IP.":".strval(LISTEN_PORT)."\n";

$loop->run();
```

Is there any possibility to trigger a kind of "garbage collection" to kill all those zombie processes? I think this does match the stated issue because the parent process (the php wrapper) does not clear the output of the child process?

PR #249, opened 2019-01-04
Fix Symfony Process deprecation warning

jasonlbeggs:
This PR just fixes the following deprecation warning:
screen shot 2019-01-04 at 10 41 52 am


freekmurze:
Thanks, I'll update composer.json to fix the tests using lower dependencies

PR #267, opened 2019-03-05
Write options to file

rubenvanassche:
This PR adds the ability to write the options given in Browsershot to a file instead of passing them as command line argument. When passing an options JSON to puppeteer that's too big(e.g. when you have an image as a header) the command line will crash because of a character overflow. We solve this by writing the options to a temporary file and passing the file as an argument instead of the whole JSON dump.

### DIFF
--- a/bin/browser.js
+++ b/bin/browser.js
@@ -4,6 +4,12 @@ const URL = require('url').URL;
 
 const [, , ...args] = process.argv;
 
+/**
+ * There are two ways for Browsershot to communicate with puppeteer:
+ * - By giving a options JSON dump as an argument
+ * - Or by providing a temporary file with the options JSON dump,
+ *   the path to this file is then given as an argument with the flag -f
+ */
 const request = args[0].startsWith('-f ')
     ? JSON.parse(fs.readFileSync(new URL(args[0].substring(3))))
     : JSON.parse(args[0]);
--- a/bin/browser.js
+++ b/bin/browser.js
@@ -1,6 +1,12 @@
 const puppeteer = require('puppeteer');
+const fs = require('fs');
+const URL = require('url').URL;
 
-const request = JSON.parse(process.argv[2]);
+const [, , ...args] = process.argv;
+
+const request = args[0].startsWith('-f ')
+    ? JSON.parse(fs.readFileSync(new URL(args[0].substring(3))))
+    : JSON.parse(args[0]);
 
 const getOutput = async (page, request) => {
     let output;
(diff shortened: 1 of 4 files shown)


tests/BrowsershotTest.php
```diff
@@ -1005,4 +1005,16 @@ public function it_can_input_form_fields_and_post_to_ssl_and_get_the_body_html()
...
+    public function it_can_write_options_to_a_file()
+    {
+        $targetPath = __DIR__.'/temp/testScreenshot.png';
+
+        Browsershot::url('https://example.com')
+            ->writeOptionsToFile()
+            ->save($targetPath);
+
+        $this->assertFileExists($targetPath);
+    }
```
brendt: I would also add a test to assert that the PDF is actually generated with this options file

bin/browser.js
```diff
@@ -1,6 +1,12 @@
 const puppeteer = require('puppeteer');
+const fs = require('fs');
+const URL = require('url').URL;
 
-const request = JSON.parse(process.argv[2]);
+const [, , ...args] = process.argv;
```
brendt: I think this is one of the few places that could benefit from an inline comment. Now there are two ways of communication between Browsershot and Puppeteer via a Bash command, I'd list them here and explain in short what the code does. I believe this would be beneficial for future maintenance.

src/Browsershot.php
```diff
@@ -645,6 +674,17 @@ protected function getNodePathCommand(string $nodeBinary): string
         return 'NODE_PATH=`npm root -g`';
     }
 
+    protected function getOptionsCommand(string $command): string
```
freekmurze: Do we know when a command becomes too large? Could we automatically switch to the options file whenever it becomes too large?

bin/browser.js
freekmurze: Agreed

bin/browser.js
```diff
@@ -1,6 +1,12 @@
...
+
+const request = args[0].startsWith('-f ')
+    ? JSON.parse(fs.readFileSync(new URL(args[0].substring(3))))
```
freekmurze: Why is that `substring(3)` needed?

danhunsaker: To strip off the leading `-f ` on the argument, I'd assume. Bit odd to have a space embedded in a CLI argument, but `¯\_(ツ)_/¯`...

src/Browsershot.php
brendt: I don't think we can use this as an automatic fallback. We've noticed in other packages that some systems don't properly support temporary directories (I believe there were some issues in our image package with it). I think it could break people's projects if we automatically switch to using this temporary directory approach. I could be wrong, but I don't think it's worth the risk.

freekmurze: Ok, let's leave this as it is in this PR for now.

PR #269, opened 2019-03-07
Feature custom papersize unit

JonathanGuo:
The `paperSize` and `margins` functions are hard coded as `mm`. It would be convenient to allow custom units.


freekmurze:
Thanks!

PR #276, opened 2019-03-25
Add dropdown/select update functionality

nshontz:
I needed a way to change the value of a form dropdown, I noticed I could update text input fields, and so I modeled the select update on the 'types' functionality. There was already a 'selects' method so I named the function 'dropdown_select'

### DIFF
--- a/README.md
+++ b/README.md
@@ -584,6 +584,26 @@ Browsershot::url('https://example.com')
     ->save($pathToImage);
 ```
 
+#### Changing the value of a dropdown value
+
+You can change the value of a dropdown on the page (you can use this to change form select fields).
+
+```php
+Browsershot::url('https://example.com')
+    ->dropdown_select('#selector1', '100')
+```
+
+You can combine `dropdown_select`, `type` and `click` to create a screenshot of a page after submitting a form:
+
+```php
+Browsershot::url('https://example.com')
+    ->type('#firstName', 'My name')
+    ->dropdown_select('#state', 'MT')
+    ->click('#submit')
+    ->delay($millisecondsToWait)
+    ->save($pathToImage);
+```
+
 #### Writing options to file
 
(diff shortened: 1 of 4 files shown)


freekmurze:
Could you also update the readme with an example of the new functionality?

README.md
```diff
@@ -584,6 +584,26 @@ Browsershot::url('https://example.com')
...
+Browsershot::url('https://example.com')
+    ->dropdown_select('#selector1', '100')
+```
+
+You can combine `dropdown_select`, `type` and `click` to create a screenshot of a page after submitting a form:
+
+```php
+Browsershot::url('https://example.com')
+    ->type('#firstName', 'My name')
+    ->dropdown_select('#state', 'MT')
```
freekmurze: rename the method to `select`

src/Browsershot.php
```diff
@@ -146,6 +146,17 @@ public function click(string $selector, string $button = 'left', int $clickCount
         return $this;
     }
 
+    public function dropdown_select(string $selector, string $value = '')
+    {
+        $dropdown_selects = $this->additionalOptions['selects'] ?? [];
```
freekmurze: Make this variable camelCased.

README.md
nshontz: There is already a method named [select](https://github.com/spatie/browsershot/blob/master/src/Browsershot.php#L215) I'm not in love with dropdown_select, but couldn't think of a good descriptive name.

danhunsaker: `select_option`?

freekmurze: `selectOption` sounds good

nshontz: Thanks! It is done!

src/Browsershot.php
nshontz: This has also been changed.

freekmurze:
Thanks!

PR #279, opened 2019-04-12
Allow custom Chromium arguments to be passed on

varbrad:
Adds the ability to set some custom arguments that will be passed onto Chromium when it is launched.

We required the ability to do this as running Chromium on our CentOS server caused some super bizarre font rendering issues unless we passed on an explicit font hinting flag to Chromium.

Any problems, let me know.

Thanks

👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,13 +117,33 @@ Browsershot::html('Foo')
 
 ### Pass custom arguments to Chromium
 
-If you need to pass custom arguments to Chromium, use the `addChromiumArg` method.
+If you need to pass custom arguments to Chromium, use the `addChromiumArguments` method.
 
-This can be useful to fix font rendering issues on some Linux distributions (e.g. CentOS).
+The method accepts an `array` of key/value pairs, or simply values. All of these arguments will automatically be prefixed with `--`.
 
 ```php
 Browsershot::html('Foo')
-  ->addChromiumArg('--font-render-hinting=none');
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -115,6 +115,17 @@ Browsershot::html('Foo')
...
 
+### Pass custom arguments to Chromium
+
+If you need to pass custom arguments to Chromium, use the `addChromiumArg` method.
+
+This can be useful to fix font rendering issues on some Linux distributions (e.g. CentOS).
+
+```php
+Browsershot::html('Foo')
+  ->addChromiumArg('--font-render-hinting=none');
```
freekmurze: Could you rename the function to `addChromiumArguments`. Let it accept an array with key values, instead of passing a string. Don't make the user type `--` before an argument, prefix that automatically.

varbrad: Thanks for the feedback! Will make the requested changes tomorrow, all are sensible and make sense to me. 👍

varbrad: @freekmurze Have made the requested changes;

- Renamed function to `addChromiumArguments`
- Accepts an array with key/values, or simply values, or both.
- Automatically prefixes the arguments with the double dash (`--`).

Any other issues, let me know.

freekmurze:
Thanks, I'll polish it a bit myself and tag a new release 👍

varbrad:
Awesome, thanks!

Issue #285, opened 2019-05-07
addChromiumArg() Does not work


shadoWalker89:
When i do this

```php
Browsershot::html('Foo')
  ->addChromiumArg([
      'font-render-hinting' => 'none',
  ]);
```

I get

`
Fatal error: Uncaught Error: Call to undefined method Spatie\Image\Manipulations::addChromiumArg() in /public_html/browsershot/vendor/spatie/browsershot/src/Browsershot.php:435 Stack trace: #0 /public_html/browsershot/index.php(325): Spatie\Browsershot\Browsershot->__call('addChromiumArg', Array) #1 {main} thrown in /public_html/browsershot/vendor/spatie/browsershot/src/Browsershot.php on line 435
`

PR #286, opened 2019-05-07
Rename addChromiumArg() to addChromiumArguments()

shadoWalker89:
Fix the readme with the correct method name
closes https://github.com/spatie/browsershot/issues/285


freekmurze:
Thanks!

Issue #280, opened 2019-04-17
Disabling Javascript


philyboysmith:
I'm increasingly coming into the issue where GDPR popups are displayed immediately and dismissDialogs() isn't proving effective.

These all seem to be loaded via javascript. Is there someway to disable javascript in puppeteer via browsershot?

Eg
```
Browsershot::url('https://www.kerrang.com/the-news/exclusive-watch-the-official-video-for-kid-bookie-and-corey-taylors-new-track/')->save('example.pdf);
```
![example](https://user-images.githubusercontent.com/144503/56284120-7cedbb00-610b-11e9-9f7f-6fba785cd669.jpg)

atymic:
It seems that puppeteer now support disabling javascript using a flag
https://github.com/GoogleChrome/puppeteer/blob/v1.17.0/docs/api.md#pagesetjavascriptenabledenabled

I'll create a PR :)

freekmurze:
👍

PR #300, opened 2019-06-06
feat: allow javascript to be disabled

atymic:
As per #280 this add an option to the package that allows you to disable javascript when capturing pages.

Let me know if you have any issues/suggestions :smile:


freekmurze:
Thanks!

atymic:
No worries :smile:

PR #305, opened 2019-06-20
Feature/specify-cookie-domain

### LINKED ISSUE #304: useCookies method forces cookie domain to match that of target URL
The `useCookies` method determines the domain based on the target URL: `$domain = parse_url($this->url)['host'];`. I imagine this is fine in most cases. However, if you wish to take a snapshot of a website where cookies utilise wildcard domains this creates a problem. This method, therefore, needs to allow each cookie to specify it's domain (at least optionally).

```php
    public function useCookies(array $cookies)
    {
        if (! count($cookies)) {
            return $this;
        }

        $domain = parse_url($this->url)['host'];

        $cookies = array_map(function ($value, $name) use ($domain) {
            return compact('name', 'value', 'domain');
        }, $cookies, array_keys($cookies));

        $this->setOption('cookies', $cookies);

        return $this;
    }
```

diggersworld:
Allows the domain to be specified when adding cookies via the `useCookies` method.

Note that it may be necessary for a developer to specify cookies for multiple domains in the same request. For this reason, the cookies passed through this method will merge rather than overwrite the previous `cookies` entry. The alternative would be to specify the domain for each cookie as it is passed in, however, that would incur a breaking change.

Fixes #304

### DIFF
--- a/README.md
+++ b/README.md
@@ -595,6 +595,14 @@ Browsershot::url('https://example.com')
    ...
 ```
 
+You can specify the domain to register cookies to, if necessary:
+
+```php
+Browsershot::url('https://example.com')
+    ->useCookies(['Cookie-Key' => 'Cookie-Value'], 'ui.example.com')
+   ...
+```
+
 #### Clicking on the page
 
 You can specify clicks on the page.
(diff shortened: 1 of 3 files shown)


freekmurze:
Thanks!

PR #306, opened 2019-06-26
add manipulations class path within readme doc

thamerbelfkihthamer:
after this StackOverflow question, I figured out that in the readme doc, we are missing: `use Spatie\Image\Manipulations;`  in the code that resizes a screenshot.

here is the link of the question [https://stackoverflow.com/questions/56771585/using-spatie-browsershot-get-class-app-http-controllers-manipulations-not-foun/56771933#56771933](url)


freekmurze:
Going to leave it like it is for now.

PR #324, opened 2019-08-07
Fixed Readme error for ->margins() method parameter

OmathiasO:
The Readme stated, that "...custom unit to the `margins` as the fourth parameter."
But the fourth parameter is the `left` margin value. Actually. the `unit` is the fifth parameter.


freekmurze:
Thanks!

PR #341, opened 2019-10-17
Connect to remote chromium instance

vitorsemeano:
This PR adds a method to the current API, allowing the configuration of a remote chromium/chrome instance. This instance needs to be running with the following parameter: --remote-debugging-port=<port goes here, default is 9222>

I also modified browser.js to use the connect method of puppeteer. Using the parameter browserURL,  puppeteer will connect with that instance, returning a browser instance, just like using the launch method. If for some reason that instance is not running, browsershot will fallback to launching a chromium instance, continuing is normal flow.

Also i added disconnect methods to browser instance when ending the browser.js script. The page instance is also being closed so the instance is kept clean after running jobs.

I also made some tests, but part of them are commented out, because they need two chromium/chrome instances running. I added description on how to proceed to test the connection part.

I also updated the README describing the behavior of setRemoteInstance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,7 +171,7 @@ Browsershot::url('https://example.com')->save($pathToImage);
 ```
 
 #### Formatting the image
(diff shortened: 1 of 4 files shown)


freekmurze:
Thanks!

PR #350, opened 2019-11-06
Get all Cookies

futureweb:
Function to get all Cookies from requested Page using 'Network.getAllCookies'

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to `Browsershot` will be documented in this file
 
+## 3.34.0 - 2019-11-06
+
+- add 'cookie' (schnederle@futureweb.at, #350)
+
 ## 3.33.0 - 2019-10-17
 
 - allow to connect to remote chromium instance (#341)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to `Browsershot` will be documented in this file
 
 ## 3.34.0 - 2019-11-06
 
-- add 'cookie' (Andreas Schnederle-Wagner, Futureweb OG, schnederle@futureweb.at, #350)
+- add function to output cookies (#350)
 
 ## 3.33.0 - 2019-10-17
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to `Browsershot` will be documented in this file
 
 ## 3.34.0 - 2019-11-06
 
-- add 'cookie' (schnederle@futureweb.at, #350)
+- add 'cookie' (Andreas Schnederle-Wagner, Futureweb OG, schnederle@futureweb.at, #350)
 
 ## 3.33.0 - 2019-10-17
 
--- a/README.md
+++ b/README.md
@@ -506,6 +506,14 @@ Browsershot also can get the body of an html page after JavaScript has been exec
 Browsershot::url('https://example.com')->bodyHtml(); // returns the html of the body
 ```
 
+### COOKIE
+
+Browsershot also can get Cookies of an html page after JavaScript has been executed:
+
+```php
+Browsershot::url('https://example.com')->cookie(); // returns cookies of requested url
+```
+
 ### Evaluate
 
 Browsershot can get the evaluation of an html page:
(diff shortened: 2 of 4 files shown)


freekmurze:
What is the use case for this?

futureweb:
> What is the use case for this?

I'm coding a Cookie Consent Manager with Cookie Crawler for GDPR / EU Cookie Law ... so I needed the possibility to get all Cookies back ... this was the most easy way to do so ... ;-)

bin/browser.js
```diff
@@ -23,6 +23,11 @@ const getOutput = async (page, request) => {
         return output;
     }
 
+    if (request.action == 'cookie') {
```
freekmurze: Use `===` here.

futureweb: makes definitely sense ... ;-)
Added Explanation to Readme, updated Changelog and corected browser.js

Hope I got all ;-)

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to `Browsershot` will be documented in this file
 
+## 3.34.0 - 2019-11-06
+
+- add 'cookie' (Andreas Schnederle-Wagner, Futureweb OG, schnederle@futureweb.at, #350)
```
freekmurze: Please remove your personal data here.

README.md
```diff
@@ -506,6 +506,14 @@ Browsershot also can get the body of an html page after JavaScript has been exec
 Browsershot::url('https://example.com')->bodyHtml(); // returns the html of the body
 ```
 
+### COOKIE
```
freekmurze: Please move this somewhere lower in the readme with the rest of the regular function and refrain form using all caps in the title.

bin/browser.js
```diff
@@ -17,12 +17,17 @@ const request = args[0].startsWith('-f ')
 const getOutput = async (page, request) => {
     let output;
 
-    if (request.action == 'evaluate') {
+    if (request.action === 'evaluate') {
```
freekmurze: 👍

src/Browsershot.php
```diff
@@ -489,6 +489,13 @@ public function bodyHtml(): string
         return $this->callBrowser($command);
     }
 
+    public function cookie(): string
```
freekmurze: Maybe we should rename this to `getCookies` so we avoid that people might think this can be used to set cookies.

CHANGELOG.md
futureweb: According to your Contributing Guidelines: Contributions are welcome and will be fully credited.
So please tell me how to format / word this so won't have another 3-4 changes until it meets your demands ;-)

(btw - "Allow edits from maintainers" is activated)

README.md
futureweb: I styled it like the other Output Options in the Readme ... it's HTML, it's PDF ... so I named it accordingly also in CAPS ... COOKIE ... sure you want this one in another Style?

src/Browsershot.php
futureweb: would prefer getCookies too, already had it that way the first time i implemented it on my side ... but as the other Output Options are also just called by Name, not get or something like that I chanced it accordingly. Sure on this to start new "naming schema"?

README.md
danhunsaker: HTML and PDF are acronyms, so they should be capitalized in any usage except file extensions. Cookie is a word, not an acronym, and so it shouldn't be in all caps unless other non-acronyms are also in all caps, which isn't the case here.

I'm not sure Cookie is generally considered an output format, though. I mean, I see how it might be, but cookies aren't the content of the page, but rather metadata for it, and metadata is extracted separately from content. (Hence the preference for `getCookie()` elsewhere.)

futureweb:
I'm puzzled if someone here would have an Idea if it could be possible to find out by which JS Script the Cookies got set by?

futureweb:
...

futureweb:
well ... as I can't re-open it ...

futureweb:
@freekmurze - mit to re-open?

lenusch:
> @freekmurze - mind to re-open?

:-( i would be happy about having help in my projects and aswell credits are always important. 
Did you forked that and made your changes in some fork?

MaxKutsayev:
@futureweb this is an important change and a fairly basic functionality. Please open pull request

PR #352, opened 2019-11-23
composer: allow Symfony 5

TomasVotruba:


freekmurze:
Thanks!

TomasVotruba:
Btw, have you considered monorepo for Spatie?
That would be one-time job for all packages. Instead of going back and forth, after `composer update` breaks down yet for another spatie package :/

PR #362, opened 2020-01-04
Adds disableImages method to prevent images from loading

aschmelyun:
I've seen a few HTML to PDF APIs that have this functionality, so I figured I'd try and add it in. 

It uses [an example](https://github.com/puppeteer/puppeteer/blob/master/examples/block-images.js) from Puppeteer's repo to block the actual image requests, and a small loop to go through the DOM and remove all `` elements after the page has been fetched.

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,6 +313,15 @@ Browsershot::url('https://example.com')
     ->save($pathToImage);
 ```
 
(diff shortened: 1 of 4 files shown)


freekmurze:
Thanks!

PR #371, opened 2020-02-18
Update README.md

davidjr82:
Updated instructions to install in forge, in order to use Node v12 (node v8 has reached EOL).

Tested in a recently provisioned Laravel Forge server, and works fine with node v12.


freekmurze:
Thanks!

PR #390, opened 2020-04-19
Add use of WS Endpoint option

jmslbam:
First of all, thank you very much for this lib. Works like a charm. 

# Background
Are you open to introducing the option "browserWSEndpoint" while using `puppeteer.connect`. I'm using https://www.browserless.io/ and that works by settings the `browserWSEndpoint` value like:

```
const browser = await puppeteer.connect({
  browserWSEndpoint: 'wss://chrome.browserless.io/'
});
```

They don't have an way to work with the `browserURL` option so hench the PR.

# Unit test
As far as the unit test I wrote. I don't know if you would like it to be enable by default? It depends on this external service to be online. We can either comment it out like the "remoteInstance" tests or keep it like this. Your call.

I do have added a test / assets to check of the WS Endpoint is online by just doing a `assetsContains` on the url instead of trying to code this with some `curl` code or so. 
I could have set up a whole new Guzzle client and do a proper request, but it didn't matter to me if the return request was JSON. I just wanted to know if the url is online.

Let me know how you feel about this all.

Kinds regards,

Jaime Martinez

### DIFF
--- a/bin/browser.js
+++ b/bin/browser.js
@@ -36,12 +36,21 @@ const callChrome = async () => {
     let remoteInstance;
 
     try {
-        if (request.options.remoteInstanceUrl) {
+        if (request.options.remoteInstanceUrl || request.options.browserWSEndpoint ) {
+            // default options
+            let options = {
+                ignoreHTTPSErrors: request.options.ignoreHttpsErrors
+            };
+
+            // choose only one method to connect to the browser instance
+            if ( request.options.remoteInstanceUrl ) {
+                options.browserURL = request.options.remoteInstanceUrl;
+            } else if ( request.options.browserWSEndpoint ) {
+                options.browserWSEndpoint = request.options.browserWSEndpoint;
+            }
+
             try {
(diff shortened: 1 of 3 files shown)


freekmurze:
Thank you for this, but I'd rather not depending on any external services.

jmslbam:
Hi Freek, thanks for the quick reply. Isn't the `remoteInstanceUrl` not also a external services? 
You can host you own instance of Browserless so you do depend on a external service.

That I use https://www.browserless.io instead of a self hosted version of this is just the case for the unit test. 

The WSEndpoint it a part of Puppeteer to connect the a Chrome instance, right? So at least use that part of the code and I will comment out the code in the PHPUnit test just like the `remoteInstanceUrl` test so it doesn't depend on a external service.

Ciao

freekmurze:
After re-reading the code, I've reconsider my stance. I'll merge this in. Thanks!

jmslbam:
Great, you're welcome! Have a good day!

PR #395, opened 2020-05-02
Libgbm-dev dependency for puppeteer 3.0

patrickbrouwers:
Browsershot seems to be working great with puppeteer 3.0. I only had to install the `ibgbm-dev` dependency after upgrading from 2.0 to 3.0. Updated the readme to reflect the need of that new dependency.


freekmurze:
Thank you!

PR #402, opened 2020-06-15
[feature] Get the list of triggered requests

rennokki:
Each request made to a website can trigger more requests, like loading the CSS files & everything. You are able to block certain ones, but you can't get a list of them.

This PR adds a new `triggeredRequests()` function that outputs a JSON string with all of the loaded URLs.

~~It can later be transformed into an array as stated in the readme.~~

Apart from `url`, I don't know what other data might be needed and that can be added to the list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,20 @@ Browsershot also can get the body of an html page after JavaScript has been exec
 Browsershot::url('https://example.com')->bodyHtml(); // returns the html of the body
 ```
 
+If you wish to retrieve a JSON list of all of the requests that the page triggered you can do so:
+
+```php
+$response = Browsershot::url('https://example.com')->triggeredRequest();
+
+$requests = json_decode($response, true);
+
+foreach ($requests as $request) {
+    $url = $request['url']; //https://example.com/
+}
+```
+
+**`triggeredRequests()` works well with `waitUntilNetworkIdle` as described [here](#waiting-for-lazy-loaded-resources)**
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -35,6 +35,20 @@ Browsershot also can get the body of an html page after JavaScript has been exec
 Browsershot::url('https://example.com')->bodyHtml(); // returns the html of the body
 ```
 
+If you wish to retrieve a JSON list of all of the requests that the page triggered you can do so:
+
+```php
+$response = Browsershot::url('https://example.com')->triggeredRequest();
```
freekmurze, line 41: Could you rename the method to `triggeredRequests()` and let it return an array?

rennokki: There you have it.

src/Browsershot.php
```diff
@@ -556,6 +556,15 @@ public function evaluate(string $pageFunction): string
         return $this->callBrowser($command);
     }
 
+    public function triggeredRequests(): array
+    {
+        $command = $this->createTriggeredRequestsListCommand();
+
+        $list = @json_decode($this->callBrowser($command), true);
```
freekmurze, line 563: Do we need that `@` there?

rennokki: Used it to suppress any non-JSON formatted output, but now I think that throwing an exception gives more information about what happens rather than showing an empty array.

freekmurze:
Thanks!

PR #406, opened 2020-07-07
Account for the removal of require('puppeteer/DeviceDescriptors')

egyptik:
Puppeteer v5.0.0 has removed the top-level file `puppeteer/DeviceDescriptors` from the Puppeteer package ([#6043](https://github.com/puppeteer/puppeteer/pull/6043)), causing the following error when using the `->device()` option:

`Error: Cannot find module 'puppeteer/DeviceDescriptors'`

The same data is exposed via `puppeteer.devices`, so `bin/browser.js` should be updated accordingly.

### DIFF
--- a/bin/browser.js
+++ b/bin/browser.js
@@ -128,7 +128,7 @@ const callChrome = async () => {
         }
 
         if (request.options && request.options.device) {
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks!

PR #411, opened 2020-07-21
Replace emulateMedia call with emulateMediaType

edcoreweb:
The emulateMedia method was removed in https://github.com/puppeteer/puppeteer/pull/6084 (version 5.0)
emulateMedia was an alias of emulateMediaType

### DIFF
--- a/bin/browser.js
+++ b/bin/browser.js
@@ -134,7 +134,7 @@ const callChrome = async () => {
         }
 
         if (request.options && request.options.emulateMedia) {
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks!

muskan3006:
Even with emulatemediatype image is not available in pdf

vitojuliano:
thanks for the info @edcoreweb

anupamdungdung:
Thanks for the info

Issue #409, opened 2020-07-12
Json parse error on windows after adding addScriptTag


waleedmazhar:
After adding js codes `->setOption('addScriptTag', json_encode(['content' => 'alert("Hello World")']))`

I am getting below error. 

```
Exit Code: 1(General error)

Working directory: C:\xampp\htdocs\abc

Output:
================


Error Output:
================
undefined:1
{"url":"https:\/\/www.abc.com\/vantage-il\/6t33g0v\/","action":"content","options":{"args":["--no-sandbox","--proxy-server=xxx.xxx.xxx.net:9000","--lang=en-US,en;q=0.9","--headless=false"],"viewport":{"width":1366,"height":768},"addScriptTag":"{\content\:\alert(\\Hello World\\)\}","executablePath":"C:\\Program Files (x86)\\Google\\Chrome\\Application\\chrome.exe","extraHTTPHeaders":{"Accept-Language":"en"},"userAgent":"Mozilla\/5.0 (X11; Linux x86_64) AppleWebKit\/537.36 (KHTML, like Gecko) Chrome\/80.0.3987.149 Safari\/537.36","waitUntil":"networkidle0","timeout":120000,"delay":1000}}
                                                                                                                                                                                                                                                                                                 ^

SyntaxError: Unexpected token c in JSON at position 289
    at JSON.parse (<anonymous>)
    at Object.<anonymous> (C:\xampp\htdocs\goldairusa\wp-content\plugins\smr-reviews\vendor\spatie\browsershot\bin\browser.js:15:12)
    at Module._compile (internal/modules/cjs/loader.js:1138:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1158:10)
    at Module.load (internal/modules/cjs/loader.js:986:32)
    at Function.Module._load (internal/modules/cjs/loader.js:879:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
    at internal/main/run_main_module.js:17:47
"
```

This is my browsershot codes
```
$a = Browsershot::url($link)
                ->setOption('addScriptTag', json_encode(['content' => 'alert("Hello World")']))
                ->setOption('args', ['--lang=en-US,en;q=0.9', '--headless=false'])
                ->setNodeBinary('C:\nodejs\node.exe')
                ->setNpmBinary('C:\nodejs\node_modules\npm\bin\npm.cmd')
[...]

PR #432, opened 2020-10-17
Workaround escaping problem on Windows by putting command into tempor…

paulrrogers:
…ary file.

Resolve https://github.com/spatie/browsershot/issues/409 by putting command into a temporary file

### DIFF
--- a/src/Browsershot.php
+++ b/src/Browsershot.php
@@ -29,6 +29,7 @@ class Browsershot
     protected $timeout = 60;
     protected $url = '';
     protected $additionalOptions = [];
+    protected $temporaryCommandDirectory;
     protected $temporaryOptionsDirectory;
     protected $writeOptionsToFile = false;
     protected $chromiumArguments = [];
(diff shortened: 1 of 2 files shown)


freekmurze:
Having to create a file for this feels like a "heavy" solution. Is there no way to do this without having to write something to the filesystem?

paulrrogers:
@freekmurze I agree this is hacky but couldn't find any other way after trying 10 different ways. Perhaps someone with deeper knowledge of the Microsoft command line and Node escaping could find a solution, though it may require changes in those dependencies.

freekmurze:
Please fork the package and use your solution there. I hope your understand that I don't want to maintain the hacky solution.

Thanks for your work on this though.

paulrrogers:
@freekmurze I understand. Thanks for your consideration. If it's not too much trouble please mark the PR as hacktoberfest-accepted.

@waleedmazhar you're free to use this PR in a fork if you like. I don't plan on maintaining a fork for my purposes.

thijsvdanker:
I understand that this will not be merged, but @paulrrogers your solution does work, thanks!

PR #448, opened 2020-11-16
Adding ability to pass envars to browser instance

badams:
This PR modifies `bin/browser.js` to to allow passing arbitrary environment variables to the browser instance when launching.

Users can now use `setOption('env', ['TZ' => 'FOO'])` to set specific environment variables which are visible to the browser.

**Why**
For our specific use case we wanted to set the timezone of the browser to match the user, I'm sure there will be other use-cases for this functionality beyond timezones.

**Tests** 
Appears it's not currently possible to test `browser.js` in this manner, but a test has been added for the php implementation of `setEnvironmentOptions`.

**API Changes**
A new method was added to the Browsershot class 

```php
$instance->setEnvironmentOptions([
    'FOO' => 'BAR',
]);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -789,7 +789,7 @@ If you want to set custom environment variables which affect the browser instanc
 
 ```php
 Browsershot::url('https://example.com')
-   ->setOption('env.TZ', 'Pacific/Auckland')
+   ->setEnvironmentOption(['TZ' => 'Pacific/Auckland'])
    ...
 ```
 
--- a/README.md
+++ b/README.md
@@ -783,6 +783,17 @@ Browsershot::url('https://example.com')
    ...
 ```
 
+#### Passing environment variables to the browser
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -783,6 +783,17 @@ Browsershot::url('https://example.com')
    ...
 ```
 
+#### Passing environment variables to the browser
+
+If you want to set custom environment variables which affect the browser instance you can use:
+
+```php
+   ->setOption('env.TZ', 'Pacific/Auckland')
```
freekmurze, line 792: Rename this to `setEnvironmentOption`. 

Also add `setEnvironmentOptions` that accepts an array with options.

freekmurze (changes requested):
Nice! Let's improve DX a little bit.

freekmurze:
Thanks!

PR #399, opened 2020-06-11
`browser.js` export `callChrome()`

ilmiont:
This PR modifies `bin/browser.js` to export `callChrome()` and enable passing in a puppeteer instance.

This allows you, when using `setBinPath()`, to create your own minimal script which configures a puppeteer instance and then calls the standard `callChrome()` with it to use the rest of Browsershot's default implementation.

**Why?**

To make it easier to use Browsershot with tools like [puppeteer-extra](https://www.npmjs.com/package/puppeteer-extra) without having to completely write your own implementation of `bin/browser.js`.

Here's an example from one of my projects:

`src/example_php.php`

```php
//...
$bs = Browsershot::url("index.html");
$bs -> addChromiumArguments(["no-sandbox"]);
$bs -> setBinPath(dirname(__DIR__) . "/libs/browsershot.js");
//...
```

`libs/browsershot.js`

```javascript
const path = require("path");
const puppeteer = require("puppeteer-extra");
const puppeteerAdblock = require("puppeteer-extra-plugin-adblocker");
const puppeteerStealth = require("puppeteer-extra-plugin-stealth");

const dir = path.dirname(__dirname);
const script = `${dir}/vendor/spatie/browsershot/bin/browser.js`;
const browser = require(script);

puppeteer.use(puppeteerAdblock({blockTrackers: true}));
puppeteer.use(puppeteerStealth());
browser.callChrome(puppeteer);
```

Note that now only a few lines of JavaScript is required to setup a custom puppeteer instance which can then be passed to Browsershot - no need to re-implement anything else within `bin/browser.js`.

**Tests** - can't see any for the relevant file...?

**API changes** - No breaking API changes.

PR #461, opened 2020-12-22
Use local require for puppeteer

spaceemotion:
I am using an approach similar to https://github.com/spatie/browsershot/pull/399#issuecomment-700730688 to run Browsershot on Vapor.

While `callChrome()` being exposed helps out a lot already, it still depends on the `puppeteer` library being installed.
Since `chrome-aws-lambda` uses `puppeteer-core`, browser.js crashes when being executed, even if you pass a custom instance.

This PR moves the actual require call into the function, so it will only be loaded when necessary.
There was also one extra line where the wrong instance was being used.

### DIFF
--- a/bin/browser.js
+++ b/bin/browser.js
@@ -1,4 +1,3 @@
-const puppeteer = require('puppeteer');
 const fs = require('fs');
 const URL = require('url').URL;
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks!

PR #468, opened 2021-01-08
Introduce a selectorIndex to bypass querySelector restrictions.

evrpress:
I would like to get the screenshot of the nth element on a page but `document.querySelector` cannot handle `eq:(xx)` like

    document.querySelector('body:eq(0)')

The solution would be to use 

    document.querySelectorAll('body')[0]

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,7 +227,7 @@ Browsershot::url('https://example.com')
     ->save($pathToImage);
 ```
 
-You can take a screenshot of an element matching a selector using `select` and an optional `$selectorIndex`.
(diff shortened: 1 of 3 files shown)


freekmurze:
The tests seem to be failing for this one. Could you take a look at that?

evrpress:
Hey @freekmurze, can you help me on the latest test please?

README.md
```diff
@@ -227,11 +227,11 @@ Browsershot::url('https://example.com')
     ->save($pathToImage);
 ```
 
-You can take a screenshot of an element matching a selector using `select`.
+You can take a screenshot of an element matching a selector using `select` and an optional `$selectorIndex`.
```
freekmurze, line 230: Could you explains what `selectorIndex` does?

freekmurze:
Don't worry about those styling issues, those will be fixed automatically. Could you improve the docs with a small explanation?

I'll merge after that is done.

freekmurze:
Thanks!

PR #495, opened 2021-04-09
Bump temporary-directory to version 2.0

bberlijn:
We cannot use Mailcoach together with this package because Mailcoach requires 2.0.

Just like https://github.com/spatie/laravel-sitemap/pull/379 ;)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
     "require": {
         "php": "^7.4|^8.0",
         "spatie/image": "^1.5.3",
-        "spatie/temporary-directory": "^1.1",
+        "spatie/temporary-directory": "^2.0",
         "symfony/process": "^5.0"
     },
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks!

danijelk:
Breaking change when on php7.4, should have been tagged as 3.4

freekmurze:
@danijelk Is it? v1 of temporary-directory is still allowed.

danijelk:
> @danijelk Is it? v1 of temporary-directory is still allowed.

In a way, we have an edge case sadly, locally running php8 it will install v2.0 of temp-dir, but in production it runs 7.4, so the lock file is invalid. Just have to be careful for now until prod is up on 8 :)

freekmurze:
Check out composers "platform" config option.

PR #496, opened 2021-04-12
feat: add ability to make POST requests

masuraoru:


src/Browsershot.php
```diff
@@ -716,6 +728,12 @@ protected function createCommand(string $url, string $action, array $options = [
 
         $command['options']['args'] = $this->getOptionArgs();
 
+        if (! empty($this->postParams) &&
```
freekmurze, line 731: This conditional is unreadable to me. Could you extract this to a function that returns a boolean. The function should use early returns to break up the conditional

README.md
```diff
@@ -714,6 +714,17 @@ Browsershot::url('https://example.com')
...
 ```
 
+#### Sending POST requests
+
+By default, all requests sent using GET method. You can make POST request to the given url by using the `postParams` method.
+Note: POST request sent using `application/x-www-form-urlencoded` content type.
+
+```php
+    ->postParams(['foo' => 'bar'])
```
freekmurze, line 724: Let's use the method `post` which is a bit friendlier. Default argument of the function should be an empty array.

freekmurze (changes requested):
Looks good, let's polish it a bit more.

src/Browsershot.php
masuraoru: I am sorry for not following your recommendations. I rethought the logic and figured out how to simplify it, at the same time adding the necessary, in my opinion, check for the presence of a URL scheme.

src/Exceptions/CouldNotTakeBrowsershot.php
```diff
@@ -15,4 +15,19 @@ public static function outputFileDidNotHaveAnExtension(string $path)
     {
         return new static("The given path `{$path}` did not contain an extension. Please append an extension.");
     }
+
+    public static function postParamsArrayIsNotAssoc()
```
freekmurze, line 19: Let's not use abbreviations, write `associative` in full here.

src/Exceptions/CouldNotTakeBrowsershot.php
```diff
@@ -15,4 +15,19 @@ public static function outputFileDidNotHaveAnExtension(string $path)
...
+    {
+        return new static('The given postParams array is not associative.');
```
freekmurze, line 21: Tack on "Make sure that every element has a key." to this message.

tests/BrowsershotTest.php
```diff
@@ -977,6 +977,31 @@ public function it_can_send_cookies()
         ], $command);
     }
 
+    /**
```
freekmurze, line 980: Make this annotation use one line, like the rest of the tests

src/Browsershot.php
```diff
@@ -113,6 +114,17 @@ public function setChromePath(string $executablePath)
         return $this;
     }
 
+    public function post(array $postParams = [])
+    {
+        if (array_keys($postParams) === range(0, count($postParams) - 1)) {
```
freekmurze, line 119: What is being checked here? Put the result in a descriptive variable like `$arrayIsFullyKeyed`.

src/Browsershot.php
```diff
@@ -716,6 +728,21 @@ protected function createCommand(string $url, string $action, array $options = [
 
         $command['options']['args'] = $this->getOptionArgs();
 
+        if (parse_url($url, PHP_URL_SCHEME) === null) {
```
freekmurze, line 731: Previously all code would work without this check. Is this extra check necessary?

src/Browsershot.php
```diff
@@ -716,6 +728,21 @@ protected function createCommand(string $url, string $action, array $options = [
...
+            throw CouldNotTakeBrowsershot::urlHasNoScheme();
+        }
+
+        if (! empty($this->postParams)) {
```
freekmurze, line 735: You could opt to extract this logic to its own function with a descriptive name.

freekmurze (changes requested):
Functionality looks good 👍
I think the code could be polished just a bit more.

src/Browsershot.php
masuraoru: Agreed, I think I shouldn't have to add it. ProcessFailedException are more than enough. I'll remove it.

src/Browsershot.php
masuraoru: As I understand now, we can safely remove this check at all, since there is nothing technically wrong with such query as '0=foo&1=bar'. But thinking for the user is often superfluous. So it was my overthinking.

tests/BrowsershotTest.php
masuraoru: Done.

src/Exceptions/CouldNotTakeBrowsershot.php
masuraoru: Removed this exception as redundant.

src/Browsershot.php
masuraoru: Now it's clean as a whistle. I apologize that my overengineering took so much of your time. =)

freekmurze:
Thanks!

PR #512, opened 2021-05-23
Add base64pdf method

MillerAdulu:
# Background
This package is platform agnostic, as such, functionality provided by the Laravel framework such as file uploads becomes a tricky issue on environments like Heroku where files can't be created on its ephemeral storage. Therefore, to use this in such an environment proves almost impossible. 

# Solution
This PR introduces a `base64pdf()` method. This method allows for a PDF to be created and returned as a base64 string, just like the `base64Screenshot` method. With the base64string, you can be able to use that to create a file using the laravel Storage facade which will persist it to your disk of choice keeping it in memory. Alternatively, using the Spatie Laravel Media Library, you can directly upload the base64 string to your disk of choice. 

# Example
## Laravel Storage
```php
    $base64_pdf = Browsershot::url('https://example.com')->base64pdf();
    @list($type, $file_data) = explode(';', $base64_pdf );
    @list(, $file_data) = explode(',', $file_data); 
    $img_name= 'cool-file-name.png';   
    Storage::disk('s3')->put($img_name, base64_decode($file_data));
```

## Laravel Media Library
```php
    $base64_pdf = Browsershot::url('https://example.com')
    ->base64pdf();
    $model->addMediaFromBase64($base64_pdf)
          ->usingFileName('cool-pdf-name.pdf')
          ->toMediaCollection();
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -447,6 +447,12 @@ You can also pass some html which will be converted to a pdf.
 Browsershot::html($someHtml)->savePdf('example.pdf');
 ```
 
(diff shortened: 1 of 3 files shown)


freekmurze:
Thanks!

PR #539, opened 2021-07-29
Update and fix install instructions for Forge servers

okaufmann:
Hi,

I think it's time to update the installation instructions for forge servers to use latest chrome and node versions. Especially `libxshmfence-dev` is now needed to make chrome work properly.

Home this helps ✌️


freekmurze:
Thank you! 🙌

PR #540, opened 2021-08-05
Add support for the `--user-data-dir` flag

### LINKED ISSUE #101: Load the same chrome profile again to get things cached
According to [puppeteer](https://github.com/GoogleChrome/puppeteer#default-runtime-settings) 

> Puppeteer creates its own Chromium user profile which it cleans up on every run.

Is there any way to load the same chrome profile again, so it should ensure that images, js & other similar things are cached.

grantholle:
There are multiple ways to skin this cat, as it is a direct option for the puppeteer [launch options](https://pptr.dev/#?product=Puppeteer&version=v10.2.0&show=api-puppeteerlaunchoptions) as `userDataDir`, but could also just be set with the `args` option. I chose using the `args` option as it seemed less invasive. This theoretically should fix #101.

### DIFF
--- a/README.md
+++ b/README.md
@@ -426,6 +426,16 @@ $image = Browsershot::url('https://example.com')
     ->screenshot()
 ```
 
+#### Setting the user data directory
+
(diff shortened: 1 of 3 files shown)


freekmurze:
Thanks!

PR #542, opened 2021-08-20
Added functionality to only send headers with navigational requests.

pharixces:
When opening pages the extra headers are sent with all subsequent request. Not every domain plays well with this strategy.

Domains with `Access-Control-Allow-Headers` will fail in preflight checks. 

I've added the option to set headers which are only send with navigational requests (e.g. the main request).

(i also converted inconsistent indenting in the browser.js file to spaces)

### DIFF
--- a/README.md
+++ b/README.md
@@ -702,6 +702,16 @@ Browsershot::url('https://example.com')
    ...
 ```
 
+#### Setting extraNavigationHTTPHeaders
+
+To only send custom HTTP headers with navigational http requests, set the extraNavigationHTTPHeaders option like so:
+
+```php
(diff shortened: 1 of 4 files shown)


freekmurze:
What is a "navigation" request?

doekenorg:
Hi @freekmurze. The navigation request is the actual url u are visiting. If that page has a css file for example, that request will get the extra headers too. These request are not navigational. 

Google fonts for example does not condone any other extra headers. So if that url uses a font of google fonts, chances are it's not getting rendered properly.

freekmurze:
Thank you

PR #564, opened 2021-09-20
Fix blockDomains and blockUrls methods

clementmas:
I found 3 issues with these methods:
- `Browsershot.php` was encoding the blocked URLs and domains twice: using `json_encode()` within the method itself and then a second time inside `getFullCommand()`. Which would generate this JSON:
```json
"blockDomains":"[\google-analytics.com\,\fonts.googleapis.com\]"
```
and cause this error:
```
SyntaxError: Unexpected token g in JSON at position 418 at JSON.parse
(Invalid escape character in string)
```
instead of generating this JSON:
```json
"blockDomains": ["google-analytics.com", "fonts.googleapis.com"]
```

- `browser.js` was referring to a variable named `domainsArray` which did not exist anymore
- `browser.js` was trying the parse the hostname from the `request` variable instead of the `interceptedRequest` variable

I added some missing tests as well.

Fixes https://github.com/spatie/browsershot/discussions/527

### DIFF
--- a/bin/browser.js
+++ b/bin/browser.js
@@ -117,18 +117,15 @@ const callChrome = async pup => {
             }
 
             if (request.options && request.options.blockDomains) {
-                const hostname = URLParse(request.url()).hostname;
-                domainsArray.forEach(function(value){
-                    if (hostname.indexOf(value) >= 0) {
-                        interceptedRequest.abort();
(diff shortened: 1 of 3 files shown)


freekmurze:
Thank you!

PR #581, opened 2021-11-24
Fix Apple Silicon Path Issue

aerni:
If you are running Browsershot on Apple Silicon, you will run into issues with your path. This PR adds the new Homebrew directory to the default path and fixes https://github.com/spatie/browsershot/discussions/580.


freekmurze:
Thanks!

ju5t:
Awesome @aerni :)

PR #589, opened 2021-12-17
Adding compatibility to Symfony 6

spackmat:
As mentioned in #588 it seems to work well also with `symfony/process` in version `6.0` and analogue to the proposed change in `spatie/image`.


freekmurze:
Thanks!

PR #610, opened 2022-03-06
Use function arguments for `mobile` and `touch`

orkhanahmadov:
Currently function arguments are not being used with `mobile()` and `touch()` function.

This change makes it possible to explicitly set options to `false` by using the function arguments.


freekmurze:
Thanks!

PR #622, opened 2022-04-02
Update savePdf to comply with 50eae92

marksalmon:
The call to `CouldNotTakeBrowsershot::chromeOutputEmpty()` now requires a second argument, since 50eae92


freekmurze:
Thank you!

PR #632, opened 2022-05-06
Hability to set initial pageNumber for templateHeaders

leonelvsc:
Based on this comment https://github.com/puppeteer/puppeteer/issues/3383#issuecomment-428613372 i've implemented and tested on my own the solution

### DIFF
--- a/bin/browser.js
+++ b/bin/browser.js
@@ -249,6 +249,25 @@ const callChrome = async pup => {
         if (request.options.delay) {
             await page.waitForTimeout(request.options.delay);
         }
+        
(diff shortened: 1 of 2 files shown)


leonelvsc:
We could use the style injection to inject css at the headerTemplate as a parameter?

freekmurze:
Could you also add an example to the docs?

leonelvsc:
Done, i will leave to anyone else the inlining of css as a parameter for headers and footers

freekmurze:
Thanks!

PR #635, opened 2022-05-19
Added method setContentUrl to set url when using html method

vitorsemeano:
This pull request adds a new method called setContentUrl.

When using the method html, sometimes you may have relative paths in your code, that needs a base url to be completely fetched. This feature allows setting this url into puppeteer.


freekmurze:
Thank you!

PR #647, opened 2022-06-27
Ability to set custom temp path

mtawil:
This PR let us set a custom temp path for those who face this issue on https://github.com/spatie/temporary-directory/issues/47


freekmurze:
Could you add a tests that proves that this works, and mention the new functionality in the readme?

PR #656, opened 2022-07-22
Add an Option to inject pagedjs polyfill

mutschler:
While paged.js works well when it's directly included in the head tag of the Site when it's added via script injection (`$browser->setOption('addScriptTag', '{"url": "https://unpkg.com/pagedjs/dist/paged.polyfill.js"}')`) there need to be done some evaluation a wait for a specific selector is added.

Im not sure about the option setter, it might be better to support the json syntax from above so one could include a local `path` version of pagedjs as well. For our case using the remote one is fine, but i could change that if you'd prefer the other variant :)

### DIFF
--- a/bin/browser.js
+++ b/bin/browser.js
@@ -297,8 +297,84 @@ const callChrome = async pup => {
             await page.addStyleTag(JSON.parse(request.options.addStyleTag));
         }
 
+        if (request.options && request.options.pagedjs) {
+            await page.addScriptTag({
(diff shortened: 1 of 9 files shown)


mutschler:
@freekmurze sorry for the ping, but is there any feedback on this? 🤔

rileysiemens:
I would also be interested in this being merged

mutschler:
@freekmurze any feedback? re-open? keep it closed?

PR #657, opened 2022-08-03
update deps

honeycube:
Hello,

Is there any option to generate image PDF using any internal method ?
I can't find any options in docs.

Maybe you have any ideas to do that ?

Thanks in advance.


honeycube:
My mistake

PR #660, opened 2022-08-03
Enable writeOptionsToFile for Windows

moisish:
If you are running browsershot on Windows, the writeOptionToFile wasn't used for the command


freekmurze:
Thanks!

PR #682, opened 2022-10-30
Security fix: Patch local file read bypass through untrusted js imports

Retr02332:
Hi Freek !

In this patch, the user can define secure URL's in a whitelist. This will prevent an attacker from embedding malicious JS from arbitrary domains.

security-fix.mp4

Regards.

### DIFF
--- a/src/Browsershot.php
+++ b/src/Browsershot.php
@@ -6,12 +6,15 @@ use Spatie\Browsershot\Exceptions\CouldNotTakeBrowsershot;
 use Spatie\Browsershot\Exceptions\ElementNotFound;
 use Spatie\Browsershot\Exceptions\FileUrlNotAllowed;
 use Spatie\Browsershot\Exceptions\HtmlIsNotAllowedToContainFile;
+use Spatie\Browsershot\Exceptions\JSLinkIsNotInAllowList;
 use Spatie\Browsershot\Exceptions\UnsuccessfulResponse;
 use Spatie\Image\Image;
 use Spatie\Image\Manipulations;
 use Spatie\TemporaryDirectory\TemporaryDirectory;
 use Symfony\Component\Process\Exception\ProcessFailedException;
 use Symfony\Component\Process\Process;
+use \DOMDocument;
+use \DOMXPath;
 
 /** @mixin \Spatie\Image\Manipulations */
 class Browsershot
@@ -58,11 +61,13 @@ class Browsershot
 
     /**
      * @param string $html
+     * @param array $allowList
      *
      * @return static
      */
-    public static function html(string $html)
+    public static function html(string $html, $allowList = [])
     {
+        (new static())->jsLinksInAllowlist($html, $allowList);
         return (new static())->setHtml($html);
     }
 
@@ -252,6 +257,26 @@ class Browsershot
         $this->proxyServer = $proxyServer;
 
         return $this;
+    }
+
+    public static function jsLinksInAllowlist(string $html, $allowList = []) {
+        // https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html#php
+        libxml_set_external_entity_loader(null);
+
+        $dom = new DOMDocument();
+        $dom->loadHTML($html);
+
+        $xpath = new DOMXPath($dom);
+
+        foreach($xpath->query("//script[@src]") as $queryResult) {
+            $jsLink = $queryResult->getAttribute("src");
+
+            foreach ($allowList as $safeDomain) {
+                if($jsLink !== $safeDomain) {
+                    throw JSLinkIsNotInAllowList::make($jsLink);
+                }
+            }
+        }
     }
 
     public function setHtml(string $html)
--- a/src/Exceptions/JSLinkIsNotInAllowList.php
+++ b/src/Exceptions/JSLinkIsNotInAllowList.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Spatie\Browsershot\Exceptions;
+
+use Exception;
+
+class JSLinkIsNotInAllowList extends Exception
+{
+    public static function make(string $jsLink)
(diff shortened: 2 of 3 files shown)


tests/BrowsershotTest.php
```diff
@@ -83,6 +84,10 @@
     Browsershot::html('<h1><img src="file://" /></h1>');
 })->throws(HtmlIsNotAllowedToContainFile::class);
 
+it('will not allow JS import links not in whitelist', function () {
```
freekmurze, line 87: Could you also add the reverse test: so add a link in the `allowList` and verify that no exception is thrown.

src/Browsershot.php
```diff
@@ -254,6 +257,26 @@ public function setProxyServer(string $proxyServer)
         return $this;
     }
 
+    public static function jsLinksInWhitelist(string $html, $safeDomains = []) {
+        // https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html#php
+        libxml_set_external_entity_loader(null);
+
+        $dom = new \DOMDocument();
```
freekmurze, line 264: Please import all namespaces so we can lose the `\`.

src/Browsershot.php
```diff
@@ -58,11 +59,13 @@ public static function url(string $url)
...
     /**
      * @param string $html
+     * @param array $safeDomains
      *
      * @return static
      */
-    public static function html(string $html)
+    public static function html(string $html, $safeDomains = [])
     {
+        (new static())->jsLinksInWhitelist($html, $safeDomains);
```
freekmurze, line 68: Right now, you're using two names for the same concept: `safeDomains` and `whiteList`. Let's make it clear by using the inclusive term `allowList` for both.

src/Browsershot.php
```diff
@@ -254,6 +257,26 @@ public function setProxyServer(string $proxyServer)
...
+        // https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html#php
+        libxml_set_external_entity_loader(null);
+
+        $dom = new \DOMDocument();
+        $dom->loadHTML($html);
+
+        $xpath = new \DOMXPath($dom);
+
+        foreach($xpath->query("//script[@src]") as $queryResult) {
+            $untrustedJSLink = $queryResult->getAttribute("src");
```
freekmurze, line 270: Rename this variable to `jsLink`

src/Browsershot.php
```diff
@@ -254,6 +257,26 @@ public function setProxyServer(string $proxyServer)
...
+
+            foreach ($safeDomains as $safeDomain) {
+                if($untrustedJSLink !== $safeDomain) {
+                    throw JSLinkIsNotWhitelisted::make();
```
freekmurze, line 274: Pass the `jsLink` to the exception and let the exception display that link, so users know which link is not allowed.

src/Browsershot.php
```diff
@@ -254,6 +257,26 @@ public function setProxyServer(string $proxyServer)
         return $this;
     }
 
+    public static function jsLinksInWhitelist(string $html, $safeDomains = []) {
+        // https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html#php
+        libxml_set_external_entity_loader(null);
```
freekmurze, line 262: Will this have side effect for people that use `DomDocument` elsewhere in the same script?

tests/BrowsershotTest.php
freekmurze, line 87: Rename all usage of `whitelist` to `allowList`.

freekmurze (changes requested):
Thanks! Let's improve it a bit!

Retr02332:
I understand !

I will be back soon with the respective changes.

Regards.

src/Browsershot.php
Retr02332: What do you mean? I can't understand this part.

Retr02332:
I have submitted the new changes.

Retr02332:
Any Update ?

src/Browsershot.php
freekmurze: Image that this code is run by someone who set this to a specific option. After running our code, that option isn't set anymore. Should we restore it to the value it was previously?

src/Browsershot.php
```diff
@@ -6,12 +6,15 @@
...
 use Spatie\Browsershot\Exceptions\FileUrlNotAllowed;
 use Spatie\Browsershot\Exceptions\HtmlIsNotAllowedToContainFile;
+use Spatie\Browsershot\Exceptions\JSLinkIsNotInAllowList;
 use Spatie\Browsershot\Exceptions\UnsuccessfulResponse;
 use Spatie\Image\Image;
 use Spatie\Image\Manipulations;
 use Spatie\TemporaryDirectory\TemporaryDirectory;
 use Symfony\Component\Process\Exception\ProcessFailedException;
 use Symfony\Component\Process\Process;
+use \DOMDocument;
```
freekmurze, line 16: Remove the unnecessary '\' here

src/Exceptions/JSLinkIsNotInAllowList.php
```diff
@@ -0,0 +1,13 @@
...
+
+namespace Spatie\Browsershot\Exceptions;
+
+use Exception;
+
+class JSLinkIsNotInAllowList extends Exception
+{
+    public static function make(string $jsLink)
+    {
+        return new static("The URL '" . $jsLink . "' is not in the allowList.");
```
freekmurze, line 11: Use string interpolation here

freekmurze (changes requested):

Retr02332:
Hey @freekmurze !

I am currently working on a critical process within the company.

When you have the merge ready, tell me to add it in the security notice.

Apologies and best regards !

PR #701, opened 2023-01-03
Docs: Use setEnvironmentOptions to set Browser Language

stefanzweifel:
It seems that a recent Chromium update broke setting the language through `setOption()` with `args`. 
(A user of mine mentioned that their custom language setting was no longer working as expected).

Apparently the proper way to set the language of the browser is to use a `LANG`-environment variable.

## Related Links

- https://bugs.chromium.org/p/chromium/issues/detail?id=755338#c14
- https://github.com/puppeteer/puppeteer/issues/1871#issuecomment-365012490
- https://github.com/puppeteer/puppeteer/issues/5970#issuecomment-690677674

### DIFF
--- a/docs/miscellaneous-options/changing-the-language-of-the-browser.md
+++ b/docs/miscellaneous-options/changing-the-language-of-the-browser.md
@@ -3,11 +3,13 @@ title: Changing the language of the browser
 weight: 3
 ---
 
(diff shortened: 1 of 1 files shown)


freekmurze:
Thank you!

PR #709, opened 2023-01-30
Fix link to Puppeteer devices in creating-images.md

kriiv:
Link to Puppeteer devices list is broken. Puppeteer repo structure has changes and removed Devices description that was previously linked. Updated link.


freekmurze:
Thank you!

PR #715, opened 2023-03-08
Add waitForSelector()

shadoWalker89:
This PR adds waitForSelector()

This puppeteer functions allows to wait for a selector to be visible in the page before printing.

This has two main use cases.

1- Wait for some content to be loaded via ajax, when that content is loaded a selector will be visible. Only then print the page

2- This can be used as an improvement to the performance.
Instead for waiting for networkidle0 or networkidle2, we can wait for a selector, once that selector is visible the print will start.
This way we don't need to wait for the connections to be closed when using networkidle0 or networkidle2

This function accepts two parameters

1- Selector which is required
2- WaitForSelectorOptions which is optional (For more info please check link below)

Puppeteer docs page for waitForSelector(): https://pptr.dev/api/puppeteer.page.waitforselector

I also added tests and updated the docs

### DIFF
--- a/bin/browser.cjs
+++ b/bin/browser.cjs
@@ -369,6 +369,10 @@ const callChrome = async pup => {
             await page.waitForFunction(request.options.function, functionOptions);
         }
 
+        if (request.options.waitForSelector) {
+            await page.waitForSelector(request.options.waitForSelector, request.options.waitForSelectorOptions ?? undefined);
+        }
+
         output = await getOutput(page, request);
 
         if (!request.options.path) {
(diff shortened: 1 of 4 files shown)


shadoWalker89:
The failing tests are related to this PR

This is the test that is failing

 it can get the failed requests

image

ahmedash95:
I need this one. thanks for adding and hopefully Spatie team merge it soon.

freekmurze:
Thanks for this!

It seems like the tests are failing and there are some merge conflicts.

I'll take a look at this PR after those things have been taken care off.

shadoWalker89:
I rebased my branch to avoid merge conflicts.
Tests on php 7.4 are failing but not due to my PR

They are failing because of this commit

https://github.com/spatie/browsershot/commit/6326adcd350137c6a8ecc7f87e7056b1c8c6b4ed

It added return types of static which is only supported on php 8.0

freekmurze:
Thanks!

PR #720, opened 2023-03-26
option to use shell_exec

syamsoul:
Hi, based on [this discussion](https://github.com/spatie/browsershot/discussions/719) (https://github.com/spatie/browsershot/discussions/719), i make this pr...

if have any better idea, i hope can make it happen.... thanks..


freekmurze:
To keep the package simple, I'm going to pass on this.

You could extend the base `Browsershot` class in your project, and override the `callBrowser` method to add your changes.

PR #730, opened 2023-05-16
Added getOutput method for full output data access on any request

vitorsemeano:
When errors occur on any method invoked by browsershot to the puppeteer browser, it's not easy to access console log data right away. Other methods exist like consoleMessages that allows access to these logs on a new browser call, not getting the original logs that were generated on the moment the first call was made.

To facilitate access to these logs, I developed the method getOuput, providing full data access after any call made to the browser.

Instead of only returning the pretended content in each browser call, I modified the output to return a standard object containing the following data:

- consoleMessages: any console messages generated during browser execution
- requestsList: all endpoints called by browser
- failedRequests: all endpoints that failed
- result: result of the method call to the browser. In case of exception this is not set
- exception: in case of a exception, the string representation of it will be here. In success this is not set

Internally this object is decomposed into the real required data by any browsershot method, and then the full object is stored and accessible using getOutput method.

If any additional call to the browser is required, the current output data will be erased and the new one will be placed accordingly.

I made two new tests, calling simple methods of browsershot, and analysing the getOutput structure in two possible situations: success and exception thrown.

### DIFF
--- a/bin/browser.js
+++ b/bin/browser.js
@@ -20,42 +20,32 @@ const consoleMessages = [];
 
 const failedRequests = [];
 
-const getOutput = async (page, request) => {
-    let output;
+const getOutput = async (request, page = null) => {
+    let output = {
+        requestsList: requestsList,
+        consoleMessages: consoleMessages,
+        failedRequests: failedRequests,
+    };
 
-    if (request.action == 'requestsList') {
-        output = JSON.stringify(requestsList);
-
-        return output;
+    if (!['requestsList', 'consoleMessages','failedRequests'].includes(request.action) && page) {
+        if (request.action == 'evaluate') {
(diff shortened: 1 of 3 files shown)


vitorsemeano:
Inspecting the automatic checks, it seems the test "can process post request" is timing out requesting the endpoint 'https://httpbin.org/post'.

This should be mocked or another approach should be used so this would not be dependent on a external service.

vitorsemeano:
I still find this relevant, as adding a easier way to get all errors, if any, from the console, while running any command.

This is very useful to debug any problem that surges, and the only way or an easier way to get some error is to get what is outputted to the console. Only seeing timeout errors is not very helpful.

I kindly request that this PR should be open again for review.

PR #735, opened 2023-05-25
Cleanup temporary html files

angelej:
Hi,
when using `Browsershot::html()` in combination with the following methods, `bodyHtml()`, `base64Screenshot()`, `screenshot()`, `base64pdf()`, `evaluate()`, `triggeredRequests()`, `consoleMessages()` or `failedRequests()`, the temporary html files never get deleted. Similar methods like `pdf()` already clean up tmp html files.

There's also a minor fix for the "can get the failed requests" test.


freekmurze:
Thanks!

PR #740, opened 2023-06-02
Fix node command error caused by type: module in package.json

osbre:
Since the official Laravel starters (Jetstream and Breeze) now have this line in `package.json`,

https://github.com/laravel/jetstream/pull/1315
```json
"type": "module",
```

This package no longer works in new Laravel projects because the `node` command execution results in an error. 

The solution would be rename the file to `cjs` which will indicate that this file runs CommonJS code.


osbre:
Thanks @timvandijck, I've updated the branch

osbre:
@timvandijck Here is a similar change made to the official Laravel package: https://github.com/laravel/octane/pull/696

ingLomeland:
@timvandijck any updates on when this can be merged?

ingLomeland:
maybe even @freekmurze could have a look? its currently broken for Laravel ^10.8 without this change

ingLomeland:
> Will this break for people on older Laravel versions?

Unknown.

osbre:
> Will this break for people on older Laravel versions?

it shouldn't unless they have a custom file that refers to the `vendor/spatie/browsershot/bin/browser.js` file.

btw, the tests are failing for unrelated reason and need to be re-run.

freekmurze:
Thanks!

PR #747, opened 2023-06-30
Read redirect history

bessone:
Sometimes it can happen that the screenshot made by Browsershot is different from the one displayed by your browser.
It is often due to one or more redirects made by the site, based on geolocation, cookies, etc...

To understand what happens, I tried to keep track of the redirects made by the site, both HTTP, Javascript and HTML ones, not using only the Puppeteer redirectChain because it only considers the HTTP ones.
I tried to capture any request that is a navigation requests that attempts to load a new document, mke sure the redirect is in the parent frame or we will see the navigation for other frames.

Since they are available, in addition to the URLs I thought of including HTTP status, reason and headers.

### DIFF
--- a/bin/browser.cjs
+++ b/bin/browser.cjs
@@ -16,6 +16,8 @@ const request = args[0].startsWith('-f ')
 
 const requestsList = [];
 
+const redirectHistory = [];
+
 const consoleMessages = [];
 
 const failedRequests = [];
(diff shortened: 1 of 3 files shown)


freekmurze:
Thank you!

freekmurze:
Could you send a PR to document this feature in our docs?

PR #756, opened 2023-07-18
Fix Windows command escaping

EmanueleCoppola:
I'm encountering a similar escaping issue on Windows as described in #681.

However, in the current version, the Windows command appears to be completely broken.
This PR addresses the problem and fixes the execution on Windows.

### DIFF
--- a/src/Browsershot.php
+++ b/src/Browsershot.php
@@ -930,6 +930,15 @@ class Browsershot
         throw new ProcessFailedException($process);
     }
 
+    protected function escapeshellarg(string $arg)
+    {
+        if ($this->isWindows()) {
+            return '"' . str_replace('"', '\\"', $arg) . '"';
+        }
+
+        return escapeshellarg($arg);
+    }
+
     protected function getFullCommand(array $command)
     {
         $nodeBinary = $this->nodeBinary ?: 'node';
@@ -940,13 +949,11 @@ class Browsershot
 
         if ($this->isWindows()) {
             $fullCommand =
-                $nodeBinary.' '
-                .escapeshellarg($binPath).' '
-                .'"'
-                .$optionsCommand
-                .'"';
+                $this->escapeshellarg($nodeBinary).' '
+                .$this->escapeshellarg($binPath).' '
+                .$optionsCommand;
 
-            return escapeshellcmd($fullCommand);
+            return $fullCommand;
         }
 
         $setIncludePathCommand = "PATH={$this->includePath}";
@@ -980,11 +987,7 @@ class Browsershot
             $command = "-f {$temporaryOptionsFile}";
         }
 
-        if ($this->isWindows()) {
-            return str_replace('"', '\"', $command);
-        }
-
-        return escapeshellarg($command);
+        return $this->escapeshellarg($command);
     }
 
     protected function arraySet(array &$array, string $key, $value): array
--- a/src/Browsershot.php
+++ b/src/Browsershot.php
@@ -964,7 +964,7 @@ class Browsershot
             $setIncludePathCommand.' '
             .$setNodePathCommand.' '
             .$nodeBinary.' '
-            .escapeshellarg($binPath).' '
+            .$this->escapeshellarg($binPath).' '
             .$optionsCommand;
     }
 


freekmurze:
Thanks!

BARNZ:
@freekmurze @EmanueleCoppola Unfortunately this commit breaks the `headerHtml` and `footerHtml` functions as they are no longer properly escaped when the command is sent through to puppeteer. All our reports are failing to generate today on windows after updating this package - a bit of triage shows this commit to be the culprit.

I've had a look at #681 but its not clear to me how specifically this commit solves that issue? We don't experience the issue listed in #681 although we do set the NODE_PATH env var and the `setNodeBinary` to our global puppeteer v20 install.

Here's a simple example comparing the command output that gets sent to puppeteer:

Current release - broken:

`
"C:\laragon\bin\nodejs\node-v18\node.exe" "C:\laragon\www\adt\vendor\spatie\browsershot\src/../bin/browser.cjs" "{\"url\":\"file:\/\/C:\\Users\\mattb\\AppData\\Local\\Temp\\611586084-0060668001690334044\\index.html\",\"action\":\"pdf\",\"options\":{\"path\":\"C:\\laragon\\www\\adt\\storage\\temp\/reports\/5d0d033d-fead-4c8d-808e-0761f4f45205\/generated.pdf\",\"args\":[],\"viewport\":{\"width\":1920,\"height\":1080},\"displayHeaderFooter\":true,\"ignoreHttpsErrors\":true,\"margin\":{\"top\":\"15mm\",\"right\":\"10mm\",\"bottom\":\"40mm\",\"left\":\"10mm\"},\"format\":\"A4\",\"headerTemplate\":\"<h1>test<\/h1>\"}}"
`

Prior release - working:

`
C:^\laragon^\bin^\nodejs^\node-v18^\node.exe ^"C:^\laragon^\www^\adt^\vendor^\spatie^\browsershot^\src/../bin/browser.cjs^" ^"^{^\^"url^\^":^\^"file:^\/^\/C:^\^\Users^\^\mattb^\^\AppData^\^\Local^\^\Temp^\^\1872440106-0693158001690334210^\^\index.html^\^",^\^"action^\^":^\^"pdf^\^",^\^"options^\^":^{^\^"path^\^":^\^"C:^\^\laragon^\^\www^\^\adt^\^\storage^\^\temp^\/reports^\/84c6cf48-1efc-4493-88af-1fd32ca07e41^\/generated.pdf^\^",^\^"args^\^":^[^],^\^"viewport^\^":^{^\^"width^\^":1920,^\^"height^\^":1080^},^\^"displayHeaderFooter^\^":true,^\^"ignoreHttpsErrors^\^":true,^\^"margin^\^":^{^\^"top^\^":^\^"15mm^\^",^\^"right^\^":^\^"10mm^\^",^\^"bottom^\^":^\^"40mm^\^",^\^"left^\^":^\^"10mm^\^"^},^\^"format^\^":^\^"A4^\^",^\^"headerTemplate^\^":^\^"^<h1^>test^<^\/h1^>^\^"^}^}^"
`

The difference being how the `headerTemplate` parameter is escaped at the end. In the broken one I get an error `The specified path is invalid.` I'm guessing because the `<` and `>` charaters are no longer escaped.

I think this commit should be reverted until a better solution for #681 is found.

src/Browsershot.php
```diff
@@ -940,13 +949,11 @@ protected function getFullCommand(array $command)
...
-                .escapeshellarg($binPath).' '
-                .'"'
-                .$optionsCommand
-                .'"';
+                $this->escapeshellarg($nodeBinary).' '
+                .$this->escapeshellarg($binPath).' '
+                .$optionsCommand;
 
-            return escapeshellcmd($fullCommand);
+            return $fullCommand;
```
BARNZ, line 956: Specifically its the change to this line here - `escapeshellcmd` ensures that html tags such as `<` and `>` are properly escaped. If I revert just this line here then everything continues to work same as before. What was the reason for removing this @EmanueleCoppola ?

BARNZ: Upon further testing it looks like `escapeshellcmd` can make a mess of the command particularly for more complex page headers and can cause json parsing to break inside browser.cjs.

I have side-stepped my issue altogether by simply using `writeOptionsToFile` for the time being.

EmanueleCoppola: After looking into it more, I realized that there are several problems when dealing with Windows escaping. One of the issues is that it doesn't work well with spaces in the paths of executable files, like "C:\Program Files\nodejs\node.exe".

To solve this problem, the best approach could be to utilize the actual Symfony Process command escaping and let it handle the escaping of paths and commands.

EmanueleCoppola:
@BARNZ I opened a pull request that should definitely fix the command escape on Windows that should also support spaces in the path like ```C:\Program Files\nodejs\node.exe```.

Could you please take a look at it #757 and see if you have any problem with your specific use case?

P.S. You're right #681 it's totally unrelated, I probably pasted the wrong discussion id.

src/Browsershot.php
BARNZ: Yep agreed!

PR #759, opened 2023-07-27
Improve security

angelej:


freekmurze:
Right now, you're checking for `file:`, shouldn't it be file:/`?
https://en.wikipedia.org/wiki/File_URI_scheme#:~:text=A%20valid%20file%20URI%20must,correct%2C%20but%20is%20often%20used.

angelej:
Hey @freekmurze ,
this was intended because the uri without a slash also works.

`Browsershot::url('file:tmp/example.txt')->bodyHtml();`
is the same as
`Browsershot::url('file:/tmp/example.txt')->bodyHtml();`

freekmurze:
I wonder if there's a more granular way. With this code people can't have `file:` in the content of their documents, and I can see people having this string in their content legit.

angelej:
There is now a stricter validation for the `Browsershot::url(string $url)`, which only allows urls with a scheme of `http://` and `https://`.
I also changed the `Browsershot::html(string $html)` filter string back to `file:/`. This won't fix the issue completely but makes it harder to exploit and less prone to false-positives. Like I said in the mail, we need another approach in order to solve this reliably.

PR #766, opened 2023-08-19
Update return types

lamberttraccard:
Add missing return types, and change the one on htmlFromFilePath from self to static to account for inheritance and child class overrides.


freekmurze:
Thanks!

PR #779, opened 2023-09-29
Fix generation of large PDFs

clementmas:
Browsershot can not currently save PDFs that are over 512 MB.

```php
Browsershot::url('https://page-containing-large-images.com')->save('large-file.pdf');
```

The issue is described quite well here: https://github.com/microsoft/playwright/issues/14675

The error `Cannot create a string longer than 0x1fffffe8 characters` occurs at `output.toString('base64')` which can not handle very large output (see [Stack Overflow thread](https://stackoverflow.com/questions/68230031/cannot-create-a-string-longer-than-0x1fffffe8-characters-in-json-parse)).

However, we don't need to convert the output to base64 when saving PDF files to disk. If we did, we should serialize the output to buffer but I don't think this is necessary.

Fixes https://github.com/spatie/browsershot/discussions/642 and https://github.com/spatie/browsershot/discussions/745.


freekmurze:
Thanks! It seems like the tests are failing, could you take a look at that?

clementmas:
The test "can evaluate page" was failing because it still needs to receive the encoded the output.

So I moved the encoding condition inside the `getOutput` function. All the tests are now passing. I should have checked earlier.

I also added `node_modules` inside .gitignore to make it easier to run `npm i puppeteer` locally before submitting PRs.

clementmas:
I adapted my PR to support the new `getOutput()` method from https://github.com/spatie/browsershot/pull/780

rw152:
Will this be merged? Seems like a helpful addition.

freekmurze:
Could you rebase this with `main`?

clementmas:
I rebased with main (v4). The tests are failing though but it's unrelated to this update.

freekmurze:
There are still some conflicts that prevent me from merging. Could you rebase again with `main`?

clementmas:
I rebased with main again. If this doesn't work, I'll open a new fresh PR.

freekmurze:
Best to open a new PR. Unfortunately, GitHub still detects conflicts, and won't let me merge this.

PR #791, opened 2023-11-16
Null coalescent operator breaks browserhot in Node < 14

erikverbeek:
We updated our Browsershot on a production server which is running a node version older than 14. This resulted in a breaking change for us.

Refactored the null coalescent operator with a ternary if operator. This restores functionality in node < 14 versions.


freekmurze:
This has been fixed by another PR. Thanks!

PR #793, opened 2023-11-20
Add firefox support

Bird87ZA:
I had quite a few issues getting Chrome working in a Docker container. After a whole weekend, I decided to rather try Firefox. It worked first run. Therefor, I thought I'd add Firefox support to the Browsershot package.

It was requested here too.
https://github.com/spatie/browsershot/discussions/712 - also adds a situation where chrome doesn't work on linux :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 [![run-tests](https://img.shields.io/github/actions/workflow/status/spatie/browsershot/run-tests.yml?label=tests&style=flat-square)](https://github.com/spatie/browsershot/actions)
 [![Total Downloads](https://img.shields.io/packagist/dt/spatie/browsershot.svg?style=flat-square)](https://packagist.org/packages/spatie/browsershot)
 
-The package can convert a webpage to an image or pdf. The conversion is done behind the scenes by [Puppeteer](https://github.com/GoogleChrome/puppeteer) which controls a headless version of Google Chrome.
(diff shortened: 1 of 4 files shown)


Bird87ZA:
ok picked up an issue with this.

The moment I set the type to jpg it tries to launch chrome. Got to fix that first.

Bird87ZA:
Nevermind, I've added more tests and they all pass. Turned out it was an issue in my code. I can push the extra tests if you want them?

freekmurze:
I wonder how our testsuite now passes. I would expect that you'd also have had to update our testing workflow to make sure that Firefox is available while running the tests on GitHub actions. 

Can you explain why the tests pass now?

Bird87ZA:
@freekmurze I wasn't aware that tests failed on Github Actions. Can you show me which tests failed?

I'll have a look at the testing workflow, to add those checks.

Bird87ZA:
Ah understood. I'll investigate and see what it does under the hood. Maybe it falls back to chrome. Because that's the weird error I got on local, it was throwing a chromium exception.

Bird87ZA:
What I did realise, is that if I do not have chromium and firefox installed, it doesn't work, so that's most likely why it works on the build.

I do not have the experience or know how with nodejs to fix this issue. I tried getting it to work, but I haven't been successful. If you'd like to close this PR, that's also fine.

freekmurze:
I'm going to pass on this for now.

Nevertheless, thank you for your work.

PR #804, opened 2023-12-20
Adding a taggedPdf option

ntaylor-86:
Adding a `taggedPdf()` method as discussed in discussion #799


freekmurze:
Thanks!

plweil:
Thanks @ntaylor-86 and @freekmurze !

PR #805, opened 2023-12-21
Allow `spatie/image v3`

lazer-hybiz:


freekmurze:
Seems like the tests are failing, could you take a look?

freekmurze:
We'll remove the dependency on image in the next version of Browsershot.

PR #813, opened 2024-01-11
Add global config for binaries and paths

aerni:
This PR adds the ability to globally configure the `nodeBinary`, `npmBinary`, `includePath`, and `nodeModulePath`. This is useful in a couple of scenarios:

1. If you are using Browsershot in many different environments and need easy access to globally change the binaries and paths. [See this discussion](https://github.com/spatie/browsershot/discussions/636).
2. When you are using a third-party package that is using Browsershot and don't have access to set the binaries and paths on the Browsershot call.
3. If you don't want to manually set the binaries and paths every time you are making a Browsershot call. It's also easy to forget that you have to manually set the binaries and paths to make things work.

I'm aware that this package is framework-agnostic. So I'm not quite sure how you'd like to handle this. Maybe creating a separate `laravel-browsershot` package is a better idea. 

Tests are currently failing because Laravel is missing in this context.

Let me know what you think.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,10 +23,11 @@
     ],
     "require": {
         "php": "^8.2",
-        "spatie/temporary-directory": "^2.0",
-        "symfony/process": "^6.0|^7.0",
+        "ext-fileinfo": "*",
         "ext-json": "*",
-        "ext-fileinfo": "*"
+        "spatie/laravel-package-tools": "^1.16",
+        "spatie/temporary-directory": "^2.0",
+        "symfony/process": "^6.0|^7.0"
     },
     "require-dev": {
         "pestphp/pest": "^1.20",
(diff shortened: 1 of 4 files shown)


freekmurze:
I'm thinking a separate package would be the way forward. Or just configure this in your own project.

edalzell:
Hey there @freekmurze, I am encountering an issue as well.

I'm using your PDF package, but locally I use `nvm` to manage node versions, but my colleagues don't, so it would be great if I could set the paths in a config or something so that my code works everywhere.

rdpascua:
I'm encountering this same issue, I have no way of overriding the browsershot inside the pdf package. I'm using `nvm` to manage `node` version as well.

edalzell:
> I'm encountering this same issue, I have no way of overriding the browsershot inside the pdf package. I'm using `nvm` to manage `node` version as well.

I ended up using a .env var and the `withBrowsershot` closure to set the path.

PR #846, opened 2024-05-23
Fixing `spatie/image`'s dependant bot alerts issues in `composer.lock`

thanosalexandris:
Having installed latest `spatie/image` version, we'll avoid dependant bot alerts.


freekmurze:
Thanks!

moisish:
@freekmurze since the ImageManipulation class is using the spatie/image, should the package be a non dev dependency?

PR #849, opened 2024-06-11
Prevent taking screenshots twice

clementmas:
[This request](https://spatie.be/docs/browsershot/miscellaneous-options/getting-failed-requests) currently takes 2 screenshots instead of 1 if there are no failed requests recorded:

```php
Browsershot::url('https://example.com')->failedRequests()
```

That's because `if ([/* empty array */])` returns false so Browsershot thinks the output hasn't been generated yet.

Checking for a non null value fixes this issue.


freekmurze:
Thanks!

PR #861, opened 2024-07-11
option to disable the capturing request

ziaratban:
Hi
This feature helps to reduce memory when using [data urls](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs).

```php
Browsershot
    ::HTML('More data urls exists in this document'')
    ->disableCaptureURLs()
;
```


freekmurze:
Thanks, could you also update the docs?

freekmurze:
Thanks!

PR #863, opened 2024-07-15
Fix - Clean up the temporary options file when browser was called successfully

Ardenexal:
## Problem
When using the option `->writeOptionsToFile()` when the PDF is successfully generated the temporary files are not removed causing the temp file to build over time eventually running out of disk space.

## Solution

When saving a PDF which internally calls the `callBrowser()` function, after retrieving the chromium result run the `$this->cleanupTemporaryOptionsFile()`.


freekmurze:
Thanks!

PR #865, opened 2024-08-05
Add options to set protocol timeout

zarulizham:
When generating large PDF, getting error:

> ProtocolError: Page.printToPDF timed out. Increase the 'protocolTimeout' setting in launch/connect calls for a higher timeout if needed.

This PR can specify `protocolTimeout()` on Browsershot. Example usage:

```php
$browsershot->protocolTimeout(60); // 60 seconds
``` 

By default the value of `protocolTimeout` is 30 seconds.


freekmurze:
Thanks!

PR #870, opened 2024-08-13
Update handling of browser output for Puppeteer ^23.0.0

bluesheep100:
Puppeteer modified how they return stream data, for browser compatibility reasons ([#12823](https://github.com/puppeteer/puppeteer/issues/12823))
After this update, a UInt8Array byte stream is returned, instead of a Buffer.

This PR adjusts the handling of this output to be compatible with v23's output. All tests appear green, save for three which fail due to the requests list not being populated. I believe I know the cause, and will submit a PR to fix it.

Tests remain green when running Puppeteer 22.15.0 as well, so this fix should be backwards compatible with older versions that returned a Buffer rather than UInt8Array.

### DIFF
--- a/bin/browser.cjs
+++ b/bin/browser.cjs
@@ -47,9 +47,14 @@ const getOutput = async (request, page = null) => {
             output.result = await page.evaluate(request.options.pageFunction);
         } else {
             const result = await page[request.action](request.options);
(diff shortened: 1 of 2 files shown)


freekmurze:
Thanks!

PR #871, opened 2024-08-13
Correctly respond to disableCaptureURLS option

bluesheep100:
This PR addresses a bug(?) introduced in [#861](https://github.com/spatie/browsershot/pull/861).

The option appears to be reacted to incorrectly. Perhaps the semantics should be inverted as well, i.e. change the option to "enableCaptureURLS" and enable it by default?

### DIFF
--- a/bin/browser.cjs
+++ b/bin/browser.cjs
@@ -159,7 +159,7 @@ const callChrome = async pup => {
         page.on('request', interceptedRequest => {
             var headers = interceptedRequest.headers();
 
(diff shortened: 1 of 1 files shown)


ziaratban:
This feature is enable by default for [BC](https://en.wikipedia.org/wiki/Backward_compatibility).
I used the word `disable` because it is near to other methods like `disableJavascript()`,`disableImages()`.

freekmurze:
Thanks!

PR #876, opened 2024-08-21
Fix empty PDF issue with Puppeteer ^23.0.0 

JeppeKnockaert:
This pull request reintroduces the changes done for Puppeteer v23, while fixing the issue introduced in https://github.com/spatie/browsershot/pull/870 resulting in a white PDF.

The issue was caused by not using base64 to communicate between puppeteer and Browsershot.

I also added a test that actually checks the contents of the PDF. To do this, I introduced the pdf-to-text package used in https://github.com/spatie/laravel-pdf and also added the same note in the README.

There is still one flaky test: 'can handle a permissions error with full output', which expects no errors, but sometimes gets a 404 on the favicon of example.com (because they have not favicon). It doesn't always fail, because the request for a favicon happens asynchronously.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -18,6 +18,9 @@ jobs:
             - name: Checkout code
               uses: actions/checkout@v4
 
+            - name: Install poppler-utils
+              run: sudo apt-get install -y poppler-utils
+
             - name: Setup PHP
               uses: shivammathur/setup-php@v2
               with:
--- a/README.md
+++ b/README.md
@@ -74,6 +74,18 @@ We highly appreciate you sending us a postcard from your hometown, mentioning wh
 
 All documentation is available [on our documentation site](https://spatie.be/docs/browsershot).
 
+## Testing
+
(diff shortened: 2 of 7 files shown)


bluesheep100:
I discovered this as well last night, but it was quite late, so I didn't have time to PR it :(
Good work though. Are you sure there's no way to write this new test without introducing another outside dependency? (pdftotext)

I'm not personally sure *why* exactly this issue manifested as it did, but as far as I could tell, attempting to return the raw buffer from the JS side would mess up the encoding or something, because the broken PDFs are encoded as UTF-8, whereas the working one (as decoded from base64 by PHP) is ANSI.

JeppeKnockaert:
Good point, I looked at it again and tried to see if I could just assert the binary data to contain the text. This works, but it also works with the empty PDF. The thing that causes it to be a white page seems to be very subtle, because all the content is actually there in the binary data.

bluesheep100:
I don't know much of anything about the PDF structure, but it appears they come out as blocks of stream data, prefixed with a length and "stream" keyword. I'd assume the difference in file encoding might affect that stream data and make it not display, while the binary remains identical.

PR #881, opened 2024-08-30
Added file:/ to the list of excluded paths and string contains & added option to disable redirects

JaredPage:
1. There are currently checks in `src/Browsershot.php` to ensure that external linking to a file does not work, as this very easily introduces a security vulnerability. 

In this PR - I simply introduce a further check to not only check for `file://` but also for `file:/` - which also works for linking to local files alongside some tests to check this.

A straightforward POC to demonstrate this being used is:
`Browsershot::url('file:/etc/password')->save("vuln.pdf");`
Which allows the password file to be generated out as a PDF. This flaw can be further exploited to bypass the security filters and read local files from the applicationserver through the use of iFrames, leading to unauthorised access and potential data compromise.

2. There exists the ability to _not_ follow redirects in puppeteer, but this functionality isn't exposed in Browsershot. Users may not want to follow redirects for security reasons as it can lead to the download of unexpected data or information leakage such as NTLM authentication material. 

In this PR - the option to disable redirects is added alongside some tests to ensure this still works when a redirect isn't returned.

https://cwe.mitre.org/data/definitions/425.html

You would have received an email from Tanto Security regarding this - who should be credited with finding these issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+## 4.3.1 - 2024-08-30
+
+### What's Changed
+
+* Add check for `file:/` URL fetching by @JaredPage in https://github.com/spatie/browsershot/pull/xyz
+* Added the ability to disable redirects via the `disableRedirects` method by @JaredPage in https://github.com/spatie/browsershot/pull/xyz
+
+**Full Changelog**: https://github.com/spatie/browsershot/compare/4.3.0...4.3.1
+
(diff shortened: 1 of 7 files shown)


JaredPage:
@freekmurze was there any update, movement or comments on this?

AlexVanderbist:
Hey @JaredPage, sorry about the delay. These changes look good, thanks for this PR!

PR #890, opened 2024-10-01
Add support for media features

alexwass-lr:
This introduces support for the [emulateMediaFeatures](https://pptr.dev/api/puppeteer.page.emulatemediafeatures) API as passing these to `emulateMediaType` is no longer supported.

It allows users to define any of the media query features [added in level 5](https://www.w3.org/TR/mediaqueries-5/), for example, enabling dark mode and reduced motion:

```php
    ->emulateMediaFeatures([
        ['name' => 'prefers-color-scheme', 'value' => 'dark'],
        ['name' => 'prefers-reduced-motion', 'value' => 'reduce']
    ])
    ->save($pathToImage);
```

### DIFF
--- a/bin/browser.cjs
+++ b/bin/browser.cjs
@@ -250,6 +250,10 @@ const callChrome = async pup => {
             await page.emulateMediaType(request.options.emulateMedia);
         }
 
+        if (request.options && request.options.emulateMediaFeatures) {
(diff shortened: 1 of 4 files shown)


AlexVanderbist:
Thanks!

PR #896, opened 2024-10-18
Fix await browser close

clementmas:
`remoteInstance` is a boolean while `disconnect()`and `close()` return promises.

The default behavior of `await remoteInstance ? browser.disconnect() : browser.close()` without parenthesis is to await the boolean which will resolve instantly.

Adding parenthesis ensures the promises are "awaited" properly.


AlexVanderbist:
Thanks!

PR #897, opened 2024-10-23
Add PDF fullPage support

Tarpsvo:
The fullPage option does not work by default when creating a PDF. This change will force the command to manually evaluate the width and height of the page before creating the PDF. It has its flaws (when document.body is not the scrollable element for example), but it will cover 99% of the cases.

Let me know if it requires changing in any way or if you're not interested at all in this.

### DIFF
--- a/src/Browsershot.php
+++ b/src/Browsershot.php
@@ -878,6 +878,13 @@ class Browsershot
             $command['options']['scale'] = $this->scale;
         }
 
+        if ($this->additionalOptions['fullPage'] ?? false) {
+            $heightPx = $this->callBrowser($this->createCommand($url, 'evaluate', ['pageFunction' => 'document.body.scrollHeight']));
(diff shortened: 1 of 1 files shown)


Tarpsvo:
Interesting... I am also on Puppeteer version 23.4.1 and it couldn't save the PDF as a single page for me, only screenshots. I'm running it on Mac locally (Node 18) and Linux (Node 20) on the server. Both displayed the same issue. Any ideas what variables am I missing here I could test?

This is currently my whole command:
```
Browsershot::url($targetPageUrl)
    ->setIncludePath('$PATH:/usr/local/bin:/opt/homebrew/bin:/usr/bin')
    ->showBackground()
    ->margins(0, 0, 0, 0)
    ->windowSize(1180, 0)
    ->fullPage()
    ->pages(1)
    ->pdf();
            ```

freekmurze:
Closing this for now as it seems to just work out of the box.

PR #907, opened 2024-12-16
Support old headless mode

clementmas:
Since Puppeteer v22, `newHeadless` is now the default. See: https://pptr.dev/guides/headless-modes

So I replaced the method `newHeadless` (which is now useless) with `oldHeadless` to use the old `chrome-headless-shell`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,10 +54,10 @@ foreach ($requests as $request) {
 }
 ```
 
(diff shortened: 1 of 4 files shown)


freekmurze:
I'm going to pass on this one for now, since changing method names is a breaking change. I'll revisit this in the future when creating a new major release.

PR #908, opened 2024-12-18
filter_var to sanitise URL

chuajianshen:
Uses PHP filter_var() function to filter the URL with the FILTER_VALIDATE_URL flag (https://www.php.net/manual/en/function.filter-var.php).


freekmurze:
Could you rebase against the current `main` branch?

freekmurze:
Thanks! Meanwhile, I added this myself on `main`. 👍

PR #911, opened 2025-01-06
Use `denylist` Chromium option to avoid including local files

freekmurze:


BoGnY:
please merge this PR
the current implementation of $unsafeProtocols used with tailwind css broke browsershot as [styling file:](https://v3.tailwindcss.com/docs/hover-focus-and-other-states#file-input-buttons) returns a false positive match

freekmurze:
I'm not sure if this PR works correctly, I only found "--chromium-deny-list" in some vague sources, not in the official docs...

PR #923, opened 2025-02-17
run-tests: disable app armor restrictions

vintagesucks:
Should fix the `No usable sandbox!` error as seen in https://github.com/spatie/browsershot/actions/runs/13367517414, via https://github.com/puppeteer/puppeteer/pull/13196. Tests still fail with this, but for different reasons.


freekmurze:
Thanks!

freekmurze:
And, ... we're green again ✅ 🥳

vintagesucks:
@freekmurze I'm not sure what you prefer, but both of the previously failing tests would have passed without any further changes if `newHeadless()` had been used in them.

PR #924, opened 2025-03-14
Fix | Check that response().frame() is not null

Dan-Hike:
This PR adds a check that `frame()` exists before checking `parentFrame()`. 
Currently the code is `response.request().frame().parentFrame()` which will throw a `TypeError: Cannot read properties of null (reading 'parentFrame')` when `frame()` is null. 
Issue raised here: https://github.com/spatie/browsershot/discussions/895
Puppeteer docs confirming `frame()` can be null: https://pptr.dev/api/puppeteer.httprequest.frame

### DIFF
--- a/bin/browser.cjs
+++ b/bin/browser.cjs
@@ -139,7 +139,8 @@ const callChrome = async pup => {
         });
 
         page.on('response', function (response) {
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks!

PR #932, opened 2025-04-28
Apply selector after wait for selector/function instead of before

itsgoingd:
Browsershot includes the ability to clip the exported image to a specific element using a `selector` option, as well as a `waitForSelector` option to wait until a certain selector appears on the page. It seems pretty useful to use these options together, eg:

```php
$browsershot
    ->waitForSelector('.export-target') // Wait until ".export-target" element appears on the page.
    ->selector('.export-target') // Only include that element in the screenshot
    ->save();
```

This is currently *not* possible, as the puppeteer script processes the `selector` option *before* `waitForSelector`.

This PR simply changes the order of how the options are processed, so the `selector` option is processed *after* `waitForSelector` and `waitForFunction`.

### DIFF
--- a/bin/browser.cjs
+++ b/bin/browser.cjs
@@ -367,6 +367,18 @@ const callChrome = async pup => {
             }, request.options.initialPageNumber);
         }
 
+        if (request.options.function) {
+            let functionOptions = {
+                polling: request.options.functionPolling,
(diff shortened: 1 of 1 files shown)


freekmurze:
Very nice, thanks!

PR #929, opened 2025-04-05
Fix: SSRF security bug

linhnph05:
I have made these changes as discussed previously in the security email:
1. Ensures that the scheme is either http or https. This ensures that no other protocols (e.g., ftp, file) are allowed
2. Host Resolution: The host is extracted from the URL. If it isn’t a valid IP address, it is resolved using gethostbyname(). This step converts domain names into their corresponding IP addresses for validation.
3. IP address validation: Validates that the resolved IP address does not fall into local, private, or reserved ranges using FILTER_FLAG_NO_PRIV_RANGE and FILTER_FLAG_NO_RES_RANGE. If it does, an exception is thrown.

I have run the tests, it passed most tests but failed 3 of them in BrowsershotTest: `Navigation Timeout` and `Class "Imagick" not found`
```shell
   FAILED  Tests\BrowsershotTest > it can get the console messages                                                                                      ProcessFailedException   
  The command "PATH=$PATH:/usr/local/bin:/opt/homebrew/bin NODE_PATH=`npm root -g` node '/Users/nguyenphanthangthong/CTF/dreamhack/web/cve/browsershot/browsershot/src/../bin/browser.cjs' '{"url":"https:\/\/bitsofco.de\/styling-broken-images\/","action":"consoleMessages","options":{"args":[],"viewport":{"width":800,"height":600}}}'" failed.

Exit Code: 1(General error)

Working directory: /Users/nguyenphanthangthong/CTF/dreamhack/web/cve/browsershot/browsershot

Output:
================


Error Output:
================
TimeoutError: Navigation timeout of 30000 ms exceeded
    at new Deferred (/opt/homebrew/lib/node_modules/puppeteer/node_modules/puppeteer-core/lib/cjs/puppeteer/util/Deferred.js:60:34)
    at Deferred.create (/opt/homebrew/lib/node_modules/puppeteer/node_modules/puppeteer-core/lib/cjs/puppeteer/util/Deferred.js:21:16)
    at new LifecycleWatcher (/opt/homebrew/lib/node_modules/puppeteer/node_modules/puppeteer-core/lib/cjs/puppeteer/cdp/LifecycleWatcher.js:69:60)
    at CdpFrame.goto (/opt/homebrew/lib/node_modules/puppeteer/node_modules/puppeteer-core/lib/cjs/puppeteer/cdp/Frame.js:148:29)
[...]

PR #934, opened 2025-05-06
Fix: SSRF security bug (refactored)

satoved:
Vulnerability: https://nvd.nist.gov/vuln/detail/CVE-2025-3192
Initial PR: https://github.com/spatie/browsershot/pull/929

What's done:
- Added a test based on https://gist.github.com/JunMing27/651998a34d57fbf71ff9d25386f1da0f
- Removed redudant protections
- Refactor the protections into ensure methods

### DIFF
--- a/docs/miscellaneous-options/access-local-resources.md
+++ b/docs/miscellaneous-options/access-local-resources.md
@@ -1,0 +1,12 @@
+---
+title: Access local resources
+weight: 28
+---
+
+By default, resources from the local network are blocked to prevent Server-Side Request Forgery (SSRF).
+
+If you need Browsershot to access private IP addresses or IPs from reserved ranges, explicitly set the flag:
+
+```php
+Browsershot::url('http://127.0.0.1', allowInternalResources: true);
+```
--- a/docs/miscellaneous-options/read-redirect-history.md
+++ b/docs/miscellaneous-options/read-redirect-history.md
@@ -1,6 +1,6 @@
 ---
 title: Read redirect history
-weight: 26
+weight: 27
 ---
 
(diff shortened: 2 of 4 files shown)


freekmurze:
The tests are failing, could you take a look?

Could you also add a method disable these new protections for people that do want to allow local files?

satoved:
@freekmurze tests seem to be failing because the https://chrome.browserless.io/ is down.

image

- Added a way to skip the protection
- Added the docs

satoved:
It's a breaking change, though. Since I defaulted to the secure option. Not sure how to handle it gracefully.

carlosmintfan:
Probably just release a major version and mark it as fix for the CVE of the last version? I don't know how lazy devs are with upgrading to a new major though lol.

carlosmintfan:
But aren't all the other fixes if bypasses also breaking changes, technically? Really depends on what the developer wants...but good to have a secure default that can be disabled 👍

linhnph05:
Hi, sorry I miss another attack vector in the PR #929 : DNS rebinding. The attacker can have their own server for example attacker.com and first, the dns will resolve to a public IP, then in the next step of the code execution, the server will resolve the IP to localhost -> The attacker can steal information etc,... I think you should consider about this attack vector before merging the pull request to prevent future CVEs. 
Here is a link may help to prevent this, hope it helps: https://cheatsheetseries.owasp.org/cheatsheets/Server_Side_Request_Forgery_Prevention_Cheat_Sheet.html.

PortableSteve:
Any movement on getting this merged in? Would love to see the vulnerability closed.

Mushoz:
So this vulnerability was simply never fixed? Snyk is still complaining about this.

PR #935, opened 2025-05-07
Fix handling of page errors with value `null`

rado-h:
Some pages may throw `null` as error. This leads to TypeError when collecting the page errors information. Set "null" as the message in such cases as there is no other meaningful value.


freekmurze:
Thanks!

PR #948, opened 2025-10-01
Enable click with non-CSS selectors

bessone:
Hello,

As I wrote in thread #947, Puppeteer allows for additional selectors beyond CSS. These can be very useful for clicks, especially those for cookie consent where the CSS classes are not static (did someone mention Facebook)?

This little PR adds a `locatorClick()` method that allows clicking on the page using non-CSS selectors (https://pptr.dev/guides/page-interactions#non-css-selectors). I thought I'd use a new method to make the change backwards compatible with those who already use `click()` without causing problems.

I'm already using it in production with success.

Thanks

Ciao

### DIFF
--- a/bin/browser.cjs
+++ b/bin/browser.cjs
@@ -337,6 +337,17 @@ const callChrome = async pup => {
             }
         }
 
+        if (request.options && request.options.locatorClicks) {
+            for (let i = 0, len = request.options.locatorClicks.length; i < len; i++) {
+                let clickOptions = request.options.locatorClicks[i];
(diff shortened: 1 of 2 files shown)


bessone:
With error handling added, the locator times out if the element being searched for doesn't exist on the page.
(https://pptr.dev/guides/page-interactions#locator-timeouts)

This way, the try/catch avoids the error, with behavior similar to `click()`, where if the element doesn't exist, nothing happens.

freekmurze:
Thanks!

PR #970, opened 2026-01-14
Deny Arbitrary Local File Read Bypasses

g4nkd:
Earlier versions allowed HTML tags to access and read arbitrary local files using payloads such as `<iframe src="\\localhost/etc/passwd">` or `<iframe src="//127.0.0.1/etc/passwd">`. This new array in "$unsafeProtocols" deny this type of exploitation.

### DIFF
--- a/src/Browsershot.php
+++ b/src/Browsershot.php
@@ -76,6 +76,8 @@ class Browsershot
         'file:\\',
         'file:\\\\',
         'view-source',
+        '\\\\',
+        '//'
     ];
 
     /** @var array<string,string> */


freekmurze:
This should be fixed in 5.2.2. The `\\` entry in the unsafe protocols list was too broad — it matched any double backslash anywhere in the HTML via `str_contains`, which causes false positives with TailwindCSS compiled output (e.g. `hover\:flex`).

The fix replaces the broad check with a targeted regex that only blocks `\\` when followed by local addresses (localhost, 127.*, etc.), same approach as the protocol-relative URL check.

slovenianGooner:
Great, will update to 5.2.2 and hopefully it works!

Thanks for a prompt response.

PR #972, opened 2026-02-09
Fix #970: Block local file read bypasses via UNC paths and protocol-relative URLs

freekmurze:
## Summary

Addresses the security issue raised in #970 (local file read via `\\localhost/...` and `//127.0.0.1/...` bypasses), but with a corrected approach:

- **UNC paths**: Added `\\` to `unsafeProtocols`, blocking HTML containing double-backslash patterns like `<iframe src="\\localhost/etc/passwd">`
- **Protocol-relative local URLs**: Added a targeted regex in `setHtml()` that blocks `//` followed by local addresses (localhost, 127.x.x.x, 0.0.0.0, [::1])

### Why not just add `//` to `unsafeProtocols`?

The original PR #970 added `//` to the `unsafeProtocols` array. This would break `setHtml()` completely because the check uses `str_contains()` — every HTML document with a URL like `https://example.com` contains `//`. The regex approach targets only local addresses while allowing legitimate protocol-relative URLs like `//cdn.example.com/style.css`.

## Test plan
- [x] Tests for UNC path payloads (`\\localhost`, `\\127.0.0.1`)
- [x] Tests for protocol-relative local URLs (`//localhost`, `//127.x.x.x`, `//0.0.0.0`, `//[::1]`)
- [x] Test that legitimate protocol-relative URLs still work
- [x] All existing tests pass (116 passed)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Browsershot.php
+++ b/src/Browsershot.php
@@ -76,6 +76,7 @@ class Browsershot
         'file:\\',
         'file:\\\\',
         'view-source',
+        '\\\\',
     ];
 
     /** @var array<string,string> */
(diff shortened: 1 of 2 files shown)


freekmurze:
This has been fixed in the latest release.

PR #973, opened 2026-03-08
Update requirements.md

colinmackinlay:
On Ubuntu 24.04, namespace restrictions cause this error:

`No usable sandbox! If you are running on Ubuntu 23.10+ or another Linux distro that has disabled unprivileged user namespaces with AppArmor`

This can be fixed by adding two lines to `/etc/syscntl.conf`.

References:
- [Laracsts](https://laracasts.com/discuss/channels/laravel/laravel-pdf-spatie-puppeteer-forge-no-usable-sandbox-error-in-production)
- [AskUbuntu](https://askubuntu.com/questions/1545324/solved-ubuntu-24-04-broke-sandboxing-how2-fix)

### DIFF
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -35,7 +35,12 @@ npx puppeteer browsers install chrome
 
 # Install dependencies
 sudo apt update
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks!

PR #975, opened 2026-04-17
feat: Use only request interception when needed

kissifrot:
## Problem

`browser.cjs` unconditionally calls `page.setRequestInterception(true)` for every request, even when none of the features that require it are used (image blocking, URL blocking, POST params, etc.).

When request interception is enabled, Puppeteer routes every network request through the Chrome DevTools Protocol, which introduces detectable timing anomalies. Anti-bot systems (Cloudflare, Datadome, etc.) fingerprint this behaviour and serve a block page instead of the real content — even when stealth plugins are active.

## Solution

Compute a `needsInterception` flag based on whether any interception-dependent option is present in the request:

   - disableImages
   - blockDomains / blockUrls
   - disableRedirects
   - extraNavigationHTTPHeaders
   - contentUrl
   - postParams

 `setRequestInterception(true) `and the `request` event listener are only registered when at least one of these is set. For a plain content/screenshot/pdf request with none of these options, the browser behaves as a native Chrome instance, making stealth evasions significantly more effective.

The `console`, `pageerror` and `response` event listeners are unaffected and remain always registered.

## Testing

I didn't feel the need to add new tests, as existing ones already cover affected code.

### DIFF
--- a/bin/browser.cjs
+++ b/bin/browser.cjs
@@ -124,8 +124,6 @@ const callChrome = async pup => {
             await page.setJavaScriptEnabled(false);
         }
 
-        await page.setRequestInterception(true);
-
         const contentUrl = request.options.contentUrl;
         const parsedContentUrl = contentUrl ? contentUrl.replace(/\/$/, "") : undefined;
         let pageContent;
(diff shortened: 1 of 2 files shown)


freekmurze (changes requested):
Thanks for tackling this — the stealth motivation is valid. However, this change regresses URL capture, which is why every CI job is failing on `it can get the requests list` (and `BrowsershotTest.php:1085` / `:1109` also assert on `getRequestsList()`).

## The regression

The `page.on('request', ...)` listener does two jobs:

1. **Populate `requestsList`** (backing `Browsershot::triggeredRequests()` and `ChromiumResult::getRequestsList()`), gated only on `!disableCaptureURLS`.
2. **Abort/modify requests** (images, blocked domains, redirects, header injection, pageContent, postParams).

This PR hides the whole listener behind `needsInterception`, so job #1 silently stops working for any caller that just wants to inspect which URLs a page loads. That's a public API surface (`triggeredRequests()` is documented) and the failing tests confirm it.

## Suggested fix

In Puppeteer, the `request` event fires regardless of `setRequestInterception(true)` — interception is only required to call `.abort()`, `.continue()`, or `.respond()`. So the two concerns can be split:

- Always register a lightweight listener that pushes to `requestsList` (unless `disableCaptureURLS`).
- Only enable `setRequestInterception(true)` and the abort/continue/respond logic when one of the truly interception-dependent options is set.

That preserves the stealth win for the common screenshot/pdf case while keeping `triggeredRequests()` intact.

## Tests

Please add a test covering the new behaviour: e.g. a plain screenshot call should still return `triggeredRequests()` correctly, and a call with `disableImages` (or another interception option) should still abort images. The existing tests prove the regression but don't pin down the intent of the change itself.

kissifrot:
Thanks for the review, I missed this part indeed :sweat: , it's now fixed
I added an additional test too :slightly_smiling_face: 

BTW if you want to test for 404 status pages you can try [httpbin](http://httpbin.org/status/404) instead of Google

PR #979, opened 2026-05-27
Fix typo in sysctl configuration instructions

intrepidws:
Corrected the file path from '/etc/syscntl.conf' to '/etc/sysctl.conf' in the requirements documentation.


freekmurze:
Thanks!

